PR #5, opened 2020-10-29
Add get schema interface for traverse app

lidongze0629:
## What do these changes do?

Add get graph schema interface for python traverse app.

## Related issue number

Fixes

### DIFF
--- a/modules/graph/fragment/arrow_fragment.h
+++ b/modules/graph/fragment/arrow_fragment.h
@@ -522,6 +522,14 @@ class ArrowFragment
     }
   }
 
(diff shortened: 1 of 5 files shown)


modules/graph/fragment/arrow_fragment.h
```diff
@@ -522,6 +522,14 @@ class ArrowFragment
     }
   }
 
+  inline label_id_t GetLabelId(const vertex_t& v) const {
```
sighingnow, line 525: We already have `vertex_label(const vertex_t& v)`.

sighingnow (approved):

PR #29, opened 2020-11-05
Prepare end-to-end testing using IO stream.

sighingnow:


sighingnow:
@andydiwenzhu  could we get this one merged to dev/io ?

PR #34, opened 2020-11-09
hive orc table support

andydiwenzhu:
## Related issue number

Fixes


modules/io/adaptors/read_hive_orc.py
```diff
@@ -0,0 +1,113 @@
...
+import pyorc
+import pyarrow as pa
+import sys
+import json
+
+from urllib.parse import urlparse
+from hdfs3 import HDFileSystem
+from vineyard.io.dataframe import DataframeStreamBuilder
+
+
```
sighingnow, line 29: I'm if it is possible to write a note here, e.g., a note about how an hive table are originaized as a batch of files under a. HDFS directory, how the hive's schema been mapped to orc's schema, the spec of orc file chunk size (by rows, by columns, or by bytes).

Such a note will be helpful for others to understand how vineyard's IO interacts with hive tables, and avoid a duplicate learning process.

sighingnow (changes requested):

PR #43, opened 2020-11-11
fix for test

andydiwenzhu:
## Related issue number

Fixes

### DIFF
--- a/modules/io/adaptors/parse_dataframe_to_bytes.py
+++ b/modules/io/adaptors/parse_dataframe_to_bytes.py
@@ -22,15 +22,30 @@ import sys
 import io
 import json
 
+from urllib.parse import urlparse
 from vineyard.io.byte import ByteStreamBuilder
 
-def parse_dataframe(vineyard_socket, stream_id, proc_num, proc_index):
+def parse_dataframe(vineyard_socket, stream_id, path, proc_num, proc_index):
     client = vineyard.connect(vineyard_socket)
     streams = client.get(stream_id)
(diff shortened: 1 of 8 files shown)


sighingnow:
Could you please take a try to add test suite for those IO adaptors?

modules/io/adaptors/parse_dataframe_to_bytes.py
```diff
@@ -58,11 +73,12 @@ def parse_dataframe(vineyard_socket, stream_id, proc_num, proc_index):
 
 if __name__ == '__main__':
     if len(sys.argv) < 5:
-        print('usage: ./parse_dataframe_to_bytes <ipc_socket> <stream_id> <proc_num> <proc_index>')
+        print('usage: ./parse_dataframe_to_bytes <ipc_socket> <stream_id> <file_path> <proc_num> <proc_index>')
```
sighingnow, line 76: I think the "parse_dataframe_to_bytes" should take a stream and yeild a stream? Why it requires a `file_path`?

modules/io/adaptors/read_hdfs_bytes.py
```diff
@@ -15,6 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import vineyard
```
sighingnow, line 18: `import vineyard` should follows the standard imports `import sys`, etc.

sighingnow (approved):
Thanks!

PR #67, opened 2020-11-25
add oss adaptor and read driver

andydiwenzhu:
## Related issue number

Fixes

### DIFF
--- a/modules/io/io/oss_io_adaptor.cc
+++ b/modules/io/io/oss_io_adaptor.cc
@@ -26,7 +26,6 @@ limitations under the License.
 #include <thread>
 #include <unordered_map>
 
-
 #include "alibabacloud/oss/OssClient.h"
 #include "alibabacloud/oss/client/RetryStrategy.h"
 #include "arrow/api.h"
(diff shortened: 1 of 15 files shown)


modules/io/io/oss_io_adaptor.cc
```diff
@@ -37,6 +39,7 @@ limitations under the License.
 
 #include "basic/ds/arrow_utils.h"
 #include "common/util/functions.h"
+#include "graph/fragment/property_graph_utils.h"
```
sighingnow, line 42: Don't depends graph module in IO.

modules/io/python/tests/test_open.py
```diff
@@ -154,3 +155,22 @@ def test_vineyard_dataframe(vineyard_ipc_socket, vineyard_endpoint, test_dataset
                      vineyard_ipc_socket=vineyard_ipc_socket,
                      vineyard_endpoint=vineyard_endpoint)
     assert filecmp.cmp('%s/p2p-31.e' % test_dataset, '%s/p2p-31.out' % test_dataset_tmp)
+
+
+def test_oss(vineyard_ipc_socket, vineyard_endpoint, test_dataset, test_dataset_tmp, oss_config):
```
sighingnow, line 160: This test case cannot be run, since there's no OSS access on github.

sighingnow (changes requested):

PR #71, opened 2020-11-27
update oss adaptor for single object parallel

andydiwenzhu:
## Related issue number

Fixes

### DIFF
--- a/modules/basic/ds/arrow_utils.h
+++ b/modules/basic/ds/arrow_utils.h
@@ -103,10 +103,10 @@ CONVERT_TO_ARROW_TYPE(float, arrow::FloatArray, arrow::FloatBuilder,
                       arrow::float32())
 CONVERT_TO_ARROW_TYPE(double, arrow::DoubleArray, arrow::DoubleBuilder,
                       arrow::float64())
-CONVERT_TO_ARROW_TYPE(RefString, arrow::StringArray, arrow::StringBuilder,
-                      arrow::utf8())
-CONVERT_TO_ARROW_TYPE(std::string, arrow::StringArray, arrow::StringBuilder,
-                      arrow::utf8())
+CONVERT_TO_ARROW_TYPE(RefString, arrow::LargeStringArray,
+                      arrow::LargeStringBuilder, arrow::large_utf8())
+CONVERT_TO_ARROW_TYPE(std::string, arrow::LargeStringArray,
+                      arrow::LargeStringBuilder, arrow::large_utf8())
 CONVERT_TO_ARROW_TYPE(arrow::TimestampType, arrow::TimestampArray,
                       arrow::TimestampBuilder,
                       arrow::timestamp(arrow::TimeUnit::MILLI))
@@ -294,8 +294,8 @@ struct EmptyTableBuilder {
       } else if (type == arrow::float64()) {
         arrow::DoubleBuilder builder;
         RETURN_ON_ARROW_ERROR(builder.Finish(&dummy));
-      } else if (type == arrow::utf8()) {
-        arrow::StringBuilder builder;
+      } else if (type == arrow::large_utf8()) {
+        arrow::LargeStringBuilder builder;
         RETURN_ON_ARROW_ERROR(builder.Finish(&dummy));
       } else if (type == arrow::null()) {
         arrow::NullBuilder builder;
--- a/modules/basic/ds/arrow_utils.h
+++ b/modules/basic/ds/arrow_utils.h
@@ -70,6 +70,15 @@ struct RefString;
     lhs = std::move(result).ValueOrDie();                     \
   } while (0)
 
+#define ARROW_CHECK_OK_AND_ASSIGN(lhs, expr)                                   \
+  do {                                                                         \
+    auto status_name = (expr);                                                 \
+    if (!status_name.ok()) {                                                   \
+      LOG(FATAL) << "Arrow check failed: " << status_name.status().ToString(); \
+    }                                                                          \
+    (lhs) = std::move(status_name).ValueOrDie();                               \
+  } while (0)
+
 template <typename T>
 struct ConvertToArrowType {};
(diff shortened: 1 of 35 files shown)


modules/io/adaptors/read_oss_dataframe.cc
```diff
@@ -43,7 +43,7 @@ int main(int argc, const char** argv) {
   VINEYARD_CHECK_OK(client.Connect(ipc_socket));
   LOG(INFO) << "Connected to IPCServer: " << ipc_socket;
 
-  std::unique_ptr<OSSIOAdaptor> oss_io_adaptor(new OSSIOAdaptor(efile.c_str()));
+  auto oss_io_adaptor = IOFactory::CreateIOAdaptor(efile);
```
sighingnow, line 46: `IOFactory` doesn't fit here, since the expection of "read_oss_dataframe" is reading oss, but if the parameter is a local file path, it doens't fail or even report proper errrors.

modules/io/adaptors/read_oss_dataframe.cc
```diff
@@ -67,13 +67,13 @@ int main(int argc, const char** argv) {
...
-  if (!st.ok()) {
-    ReportStatus("error", st.ToString());
-    VINEYARD_CHECK_OK(st);
+  if (table) {
+    auto st = writer->WriteTable(table);
+    if (!st.ok()) {
+      ReportStatus("error", st.ToString());
+      VINEYARD_CHECK_OK(st);
+    }
   }
```
sighingnow, line 76: Or? `table == nullptr` indicates an error or not?

modules/io/io/io_factory.cc
```diff
@@ -41,12 +41,10 @@ void IOFactory::Finalize() { LocalIOAdaptor::Finalize(); }
...
 std::unique_ptr<IIOAdaptor> IOFactory::CreateIOAdaptor(
     const std::string& location, Client*) {
-  std::error_code err;
-  network::uri instance(location, err);
-  std::string scheme;
-  if (!err) {
-    scheme = instance.scheme().to_string();
-  } else {
+  size_t pos = location.find_first_of(':');
+  std::string scheme = location.substr(0, pos);
```
sighingnow, line 45: Why not `network::uri`? The target of `network:uri` is to avoid ad-hoc parsing routines.

modules/io/io/oss_io_adaptor.cc
```diff
@@ -383,8 +341,63 @@ Status OSSIOAdaptor::ReadTable(std::shared_ptr<arrow::Table>* table) {
...
+  } else {
+    end_ = total_len_;
+  }
+
+  VLOG(2) << "Partial Read: " << part_id_ << " " << begin_ << " " << end_ << " "
+          << total_len_;
+  return Status::OK();
+}
+
+Status OSSIOAdaptor::getTotalSize(size_t& size) {
```
sighingnow, line 368: `getTotalSize` -> `getObjectSize`.

modules/io/io/oss_io_adaptor.cc
```diff
@@ -472,103 +484,19 @@ Status OSSIOAdaptor::ReadTable(std::shared_ptr<arrow::Table>* table) {
...
     VLOG(2) << "getObjectToBuffer success, Content-Length: " << content_length;
-    content.clear();
-    content.resize(content_length);
-    auto stream = outcome.result().Content();
-    for (int i = 0; i < content_length; ++i) {
-      stream->get(content[i]);
-    }
+    std::ostringstream ss;
+    ss << outcome.result().Content()->rdbuf();
+    content = ss.str();
```
sighingnow, line 499: That will copy the content twice, right?

modules/io/io/oss_io_adaptor.h
```diff
@@ -93,13 +95,11 @@ class OSSIOAdaptor : public IIOAdaptor {
...
-  Status listAllObjects(const std::string& prefix, const std::string& suffix);
+  Status getTotalSize(size_t& size);
 
-  void selectObjects();
+  Status readLine(std::string& line, size_t& cursor);
 
-  Status producerRoutine();
-
-  Status getObjectToBuffer(const std::string& object_name, std::string& block);
+  Status getRange(const size_t begin, const size_t end, std::string& content);
```
sighingnow, line 102: Is there other option that we could void using `std::string` to wrap the result? to avoid coping as best as we can.

sighingnow (approved):
Looks good to me, just some minor comments.

modules/io/io/io_factory.cc
andydiwenzhu: Because it fails to parse uri containing column names and types

siyuan0322: If uri lib cannot parse out uri, then it's the problem of out uri design. We can discuss how to design the oss uri.

siyuan0322: https://code.aone.alibaba-inc.com/GraphScope/libgrape-alibaba/blob/master/grape/internal/io/tunnel_io_adaptor.cc#L703
This may help.

siyuan0322 (approved):
LGTM

modules/basic/ds/arrow_utils.h
```diff
@@ -70,6 +70,15 @@ struct RefString;
...
   } while (0)
 
+#define ARROW_CHECK_OK_AND_ASSIGN(lhs, expr)                                   \
+  do {                                                                         \
+    auto status_name = (expr);                                                 \
+    if (!status_name.ok()) {                                                   \
+      LOG(FATAL) << "Arrow check failed: " << status_name.status().ToString(); \
+    }                                                                          \
+    (lhs) = std::move(status_name).ValueOrDie();                               \
+  } while (0)
```
sighingnow, line 80: Ues `CHECK_ARROW_ERROR_AND_ASSIGN`.

PR #73, opened 2020-11-29
Add some utf8 branch

siyuan0322:

### DIFF
--- a/modules/basic/ds/arrow_utils.cc
+++ b/modules/basic/ds/arrow_utils.cc
@@ -77,7 +77,7 @@ std::shared_ptr<Column> CreateColumn(
     return std::make_shared<DoubleColumn>(chunked_array, chunk_size);
   } else if (type == arrow::large_utf8()) {
     return std::make_shared<StringColumn>(chunked_array, chunk_size);
-  } else if (type == arrow::binary()) {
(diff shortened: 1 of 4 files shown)


modules/basic/ds/arrow_utils.cc
```diff
@@ -77,7 +77,7 @@ std::shared_ptr<Column> CreateColumn(
     return std::make_shared<DoubleColumn>(chunked_array, chunk_size);
   } else if (type == arrow::large_utf8()) {
     return std::make_shared<StringColumn>(chunked_array, chunk_size);
-  } else if (type == arrow::binary()) {
+  } else if (type == arrow::large_binary()) {
```
sighingnow, line 80: Could we keep the both case here?

sighingnow (approved):
Looks good to me, but I'm wondering if this PR is ready for merging?

modules/basic/ds/arrow_utils.cc
siyuan0322: StringColumn's ArrayType is declared like `ConvertToArrowType<std::string>`.

sighingnow: I see. Thanks for the clearification.

PR #74, opened 2020-11-30
Refine ThreadGroup

pwrliang:
1. Refine ThreadGroup
2. Cleanup error message
3. Fix Context protocols


modules/graph/utils/thread_group.h
```diff
@@ -31,24 +33,66 @@ class ThreadGroup {
   using return_t = Status;
 
  public:
+  explicit ThreadGroup(tid_t parallelism = std::numeric_limits<tid_t>::max())
```
sighingnow, line 36: The `parallelism` obviously cannot be `MAX_INT`,  would `std::thread::hardware_concurrency` be better?

modules/graph/loader/arrow_fragment_loader.h
```diff
@@ -155,21 +155,21 @@ class ArrowFragmentLoader {
     BOOST_LEAF_CHECK(initPartitioner());
 #if defined(WITH_PROFILING)
     double init_partitioner_ts = GetCurrentTime();
-    VLOG(2) << "initPartitioner uses " << (init_partitioner_ts - start_ts)
+    VLOG(1) << "initPartitioner uses " << (init_partitioner_ts - start_ts)
```
sighingnow, line 158: Why those `2` to `1`? If i remember it correctly, `2` is less verbose?

sighingnow (changes requested):

sighingnow (approved):

PR #75, opened 2020-11-30
Reorg arrow related operations and support generate edge id.

luoxiaojian:


modules/graph/loader/arrow_fragment_loader.h
```diff
@@ -158,27 +121,29 @@ class ArrowFragmentLoader {
...
         partial_e_tables_(partial_e_tables),
         directed_(directed),
+        generate_eid_(generate_eid),
         basic_arrow_fragment_loader_(comm_spec) {}
 
   ArrowFragmentLoader(vineyard::Client& client,
                       const grape::CommSpec& comm_spec,
                       const std::vector<std::string>& efiles,
-                      bool directed = true)
+                      bool directed = true, bool generate_eid = false)
```
sighingnow, line 138: Just a little concern about the default value, should it be false? Otherwise all graphs that been loaded into vineyard requires an extra explict argument `generate_eid=True` to make it available for maxgraph. In graphscope, we assume every propery graph could work with gremlin queries.

sighingnow (approved):

PR #77, opened 2020-12-01
Fix compile issues for gcc-4.9.2

pwrliang:


pwrliang:
Changes are included in https://github.com/alibaba/libvineyard/pull/78

PR #86, opened 2020-12-03
Fix parallel read bug in HDFS reader.

andydiwenzhu:
## Related issue number

Fixes


sighingnow:
It will be better to add a note for such confusing part.

sighingnow (approved):

PR #96, opened 2020-12-10
Use arrow compute to do general cast

siyuan0322:


sighingnow (approved):
LGTM 👍 

Mergeable when graphscope test is green based on this PR.

PR #102, opened 2020-12-18
Migrate remote object to local in vineyard.

sighingnow:


sighingnow:
Will rebase after #116 merged.

PR #108, opened 2020-12-25
Merge subtrees when persist.

### LINKED ISSUE #107: Optimize metadata management in etcd, to reduce the txn size when persisting
**Describe your problem**

Etcd doesn't do very well for many keys in a single txn, the default limitation of ops in a txn is 128, and etcdserver raises many warnings about "execute too long to execute" both in update and range query.

Meanwhile flatten metadatas in etcd doesn't bring many benefits.

**Additional context**

None

siyuan0322:
Fixes #107

### DIFF
--- a/src/server/services/meta_service.h
+++ b/src/server/services/meta_service.h
@@ -41,6 +41,17 @@ limitations under the License.
 
 namespace vineyard {
 
+namespace metatree {
+enum class NodeType {
+  Value = 0,
+  Link = 1,
+  InvalidType = 15,
+};
+
+void encode_value(NodeType type, const std::string& value, std::string& str);
+void decode_value(const std::string& str, NodeType& type, std::string& value);
+}  // namespace metatree
+
 namespace asio = boost::asio;
 
 /**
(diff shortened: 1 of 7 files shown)


src/server/services/meta_service.h
```diff
@@ -618,6 +620,42 @@ class IMetaService {
...
+        std::string encoded_value;
+        if (it->second.empty()) {
+          if (it->first == "id" || it->second.data().empty()) {
+            continue;
+          }
+          encoded_value = "v";
+          if (it->first == "transient") {
+            encoded_value += "false";
+          } else {
+            encoded_value += it->second.data();
```
sighingnow, line 643: Use `encode_value()`.

src/server/services/meta_service.h
```diff
@@ -618,6 +620,42 @@ class IMetaService {
...
+          }
+        } else {
+          encoded_value = "l";
+          encoded_value += it->second.get<std::string>("__subtree", "");
```
sighingnow, line 647: Use `encode_value()`.

src/server/util/meta_tree.cc
```diff
@@ -405,46 +406,36 @@ static void generate_put_ops(const ptree& meta, const ptree& diff,
...
-static void generate_persist_ops(const ptree& diff, const std::string& name,
+/// Store metas to etcd.
+/// We create an entry for each subtree recursively, and reset processed
+/// subtree to its link to avoid duplicate information.
+static void generate_persist_ops(ptree& diff, const std::string& name,
                                  std::vector<IMetaService::op_t>& ops,
                                  std::set<std::string>& dedup) {
   std::string key_prefix = "data." + name + ".";
-  for (ptree::const_iterator it = diff.begin(); it != diff.end(); ++it) {
+  for (ptree::iterator it = diff.begin(); it != diff.end(); ++it) {
```
sighingnow, line 416: Why it cannot be `const iterator`?

sighingnow (changes requested):
Why we need a further `__subtree` field?

src/server/util/meta_tree.cc
siyuan0322: It modified the tree inplace (delete processed subtree to avoid duplication)

src/server/services/meta_service.h
siyuan0322: It involves several class and helper functions from namespace `meta_tree` in `meta_tree.cc`.  I would like to avoid the copy-paste....

src/server/services/meta_service.h
siyuan0322: Having a `__subtree` can help to recover the link, otherwise the link would take up an additional record. This way the record number can be halved.

sighingnow (approved):

PR #115, opened 2020-12-31
Fixes compatiblity bug for pandas 1.2.0.

### LINKED ISSUE #112: pandas 1.2.0 incompatible issue
**Describe your problem**

pandas 1.2.0 includes some incompatible changes that break our builder and resolver for `DataFrame`.

**Additional context**

Also affects mars: mars-project/mars#1845

sighingnow:
## What do these changes do?

Pandas 1.2.0 drops python 3.6 support, and pandas 1.0.0 drops python 3.5 support. And there's are a breaking change in python 1.2.0 that affects us.

## Related issue number

Fixes #112

### DIFF
--- a/python/vineyard/data/dataframe.py
+++ b/python/vineyard/data/dataframe.py
@@ -16,7 +16,6 @@
 # limitations under the License.
 #
 
-from distutils.version import LooseVersion
 import json
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


python/vineyard/data/dataframe.py
```diff
@@ -51,7 +52,12 @@ def dataframe_resolver(obj, resolver):
     index_size = 0
     for idx, name in enumerate(columns):
         np_value = resolver.run(obj.member('__values_-value-%d' % idx))
-        blocks.append(Block(np.expand_dims(np_value, 0), slice(idx, idx + 1, 1)))
+        if LooseVersion(pd.__version__) < LooseVersion('1.2.0'):
+            kw = dict()
+        else:
+            # ndim: 1 for SingleBlockManager/Series, 2 for BlockManager/DataFrame
+            kw = {'ndim': 2}
+        blocks.append(Block(np.expand_dims(np_value, 0), slice(idx, idx + 1, 1), **kw))
```
siyuan0322, line 60: Why don't always pass `ndim=2`? It should work fine for both versions and don't need additional version check. When pandas < 1.2.0 the signature is `ndim=None`, and after 1.2.0 they just drop the default None.

siyuan0322: or more intelligently, use `value = np.expand_dims(np_value, 0); ndim=value.ndim`

sighingnow: Fixed. Thanks!

sighingnow:
@andydiwenzhu Any comment ?

PR #119, opened 2021-01-04
support sparse data structure of pandas serialize/deserialize

acezen:
## What do these changes do?
support sparse data structure of pandas serialize/deserialize

pyarrow not support serialize/deserialize sparse data, use pickle when the data not support in pyarrow.

### DIFF
--- a/python/vineyard/data/default.py
+++ b/python/vineyard/data/default.py
@@ -20,6 +20,10 @@ import pickle
 
 try:
     import pyarrow as pa
+    try:
+        from pyarrow.serialization import SerializationCallbackError
+    except ImportError:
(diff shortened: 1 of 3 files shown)


acezen:
cc/ @sighingnow  @andydiwenzhu

sighingnow:
Could you give an example, or a test case, to demostrate the problem?

acezen:
> Could you give an example, or a test case, to demostrate the problem?

good suggestion. I add two tests case to in latest commit.

sighingnow (approved):
LGTM. Wait for comment from @andydiwenzhu

andydiwenzhu (commented):
LGTM

PR #126, opened 2021-01-05
Conform the max-txn-ops requirement of etcd.

sighingnow:
## What do these changes do?

We don't need such flags for etcd server anymore, by splitting a large update set to many transactions. We compute etcd ops by DFS when persisting object, ensuring the correctness of segmented updates for etcd.

## Related issue number

This pull request will remedy #107 as well.


andydiwenzhu (commented):
LGTM

PR #138, opened 2021-01-08
Markstream to enable 1-producer-1-consumer on server side

### LINKED ISSUE #133: stream one producer one consumer control on server side
**Describe your problem**

stream one producer one consumer control

**Descirption of the problem**

Current one producer one consumer control is on the user's side, which is weak

**Additional context**

We would like to control this in vineyardd by marking the stream when it is open the first time

andydiwenzhu:
## Related issue number 

Fixes #133

### DIFF
--- a/modules/basic/stream/byte_stream.vineyard-mod
+++ b/modules/basic/stream/byte_stream.vineyard-mod
@@ -153,8 +153,11 @@ class ByteStream : public Registered<ByteStream> {
    * @return The unique pointer to the reader
    */
   std::unique_ptr<ByteStreamReader> OpenReader(Client& client) {
-    return std::unique_ptr<ByteStreamReader>(
-        new ByteStreamReader(client, id_, meta_));
+    if (client.MarkStream(id_, 2).ok()) {
+      return std::unique_ptr<ByteStreamReader>(
+          new ByteStreamReader(client, id_, meta_));
+    }
+    return nullptr;
   }
 
   /**
@@ -164,8 +167,11 @@ class ByteStream : public Registered<ByteStream> {
    * @return The unique pointer to the writer
    */
   std::unique_ptr<ByteStreamWriter> OpenWriter(Client& client) {
-    return std::unique_ptr<ByteStreamWriter>(
-        new ByteStreamWriter(client, id_, meta_));
+    if (client.MarkStream(id_, 1).ok()) {
+      return std::unique_ptr<ByteStreamWriter>(
+          new ByteStreamWriter(client, id_, meta_));
+    }
+    return nullptr;
   }
 
   std::unordered_map<std::string, std::string> GetParams() { return params_; }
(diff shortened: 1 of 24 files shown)


modules/basic/stream/byte_stream.vineyard-mod
```diff
@@ -153,8 +153,11 @@ class ByteStream : public Registered<ByteStream> {
...
    */
   std::unique_ptr<ByteStreamReader> OpenReader(Client& client) {
-    return std::unique_ptr<ByteStreamReader>(
-        new ByteStreamReader(client, id_, meta_));
+    if (client.MarkStream(id_, 2).ok()) {
+      return std::unique_ptr<ByteStreamReader>(
+          new ByteStreamReader(client, id_, meta_));
+    }
+    return nullptr;
   }
```
sighingnow, line 161: I suggest to have API like

```cpp
Status OpenReader(Client &client, std::unique_ptr<ByteStreamReader> &r) {
   RETURN_ON_ERROR(client.MarkStream(id_, 2))....
   r = ...
   return Status::OK();
}
```

which will be more consistent will the style of our other APIs, and more importantly, let user know why the operation fail. Return `nullptr` sliently usually the root of many unintended bugs.

src/server/memory/stream_store.h
```diff
@@ -86,6 +88,7 @@ class StreamStore {
   std::shared_ptr<BulkStore> store_;
   size_t threshold_;
   std::unordered_map<ObjectID, std::shared_ptr<StreamHolder>> streams_;
+  std::unordered_map<ObjectID, int64_t> stream_marks_;
```
sighingnow, line 91: Could we use the `mark` as an attribute of `StreamHolder`? without introduce an extra map. Considering that when deleting a stream, we need an extra deletion and may lead to inconsistency.

test/stream_test.cc
```diff
@@ -83,6 +86,9 @@ int main(int argc, char** argv) {
     CHECK(byte_stream != nullptr);
 
     auto writer = byte_stream->OpenWriter(writer_client);
+    auto failed_writer = byte_stream->OpenWriter(writer_client);
+    CHECK(failed_writer == nullptr);
```
sighingnow, line 90: For the check, you may also need to consider all places the `OpenReader/OpenWriter` are called, not merely the test case.

sighingnow (changes requested):
I prefer to rename `MarkStream` as `OpenStream`, add define an enumeration

```cpp
enum class OpenStreamMode {
   read = 1,
   write = 2,
}
```

in a newly added `stream_utils.h` or somewhere like that, then replace magic `1` and `2` in `OpenReader`/`OpenWriter` with the enum, for better code readability and maintain-ability.

sighingnow (approved):

sighingnow:
LGTM.

PR #153, opened 2021-01-21
Serialization to stream and deserialize from stream.

siyuan0322:


src/client/ds/blob.cc
```diff
@@ -128,6 +145,22 @@ void BlobWriter::AddKeyValue(std::string const& key, std::string&& value) {
...
 
+void BlobWriter::Dump() const {
+  if (VLOG_IS_ON(10)) {
+    std::stringstream ss;
+    ss << "size = " << size() << ", buffer = ";
+    {
+      std::ios::fmtflags os_flags (std::cout.flags());
+      auto ptr = reinterpret_cast<const uint8_t *>(this->data());
+      for (size_t idx = 0; idx < size(); ++idx) {
+        ss << "\\x" << std::hex << static_cast<const uint32_t>(ptr[idx]);
```
sighingnow, line 156: add `<< std::setfill('0') << std::setw(2)`.

PR #164, opened 2021-01-25
Add serialize and deserialize methods for objects.

siyuan0322:


modules/io/adaptors/deserializer.cc
```diff
@@ -0,0 +1,185 @@
...
+  target.SetId(target_id);
+
+  return Status::OK();
+}
+
+/// Restore meta and blobs from stream
+/// If is a global object, then only restore local object
+/// Return local object's id,
+/// The caller is responsible for assemble objects into a global object.
+Status Deserialize(Client& client, ObjectID stream_id, std::string* out_id) {
```
sighingnow, line 81: `std::string* out_id` could we use a reference instead?

modules/io/adaptors/read_bytes.py
```diff
@@ -41,70 +41,122 @@ def read_bytes(
...
     builder = ByteStreamBuilder(client)
 
-    header_row = read_options.get("header_row", False)
-    for k, v in read_options.items():
-        if k in ("header_row", "include_all_columns"):
-            builder[k] = "1" if v else "0"
-        elif k == "delimiter":
-            builder[k] = bytes(v, "utf-8").decode("unicode_escape")
-        else:
+    serialization_mode = read_options.pop('serialization_mode', False)
```
sighingnow, line 61: Why do we need to make `read_byte` to aware if we are doing serialization?

modules/io/python/tests/test_serialize.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+import itertools
+import numpy as np
+
+import vineyard
+import vineyard.io
+
+
+@pytest.fixture(scope='module')
+def global_obj(vineyard_ipc_socket):
```
sighingnow, line 28: Why do you need such a fixture?

python/vineyard/io/stream.py
```diff
@@ -89,6 +89,33 @@ def write(path, stream, *args, **kwargs):
...
+def deserialize(path, *args, **kwargs):
+    parsed = urlparse(path)
+    if not parsed.scheme:
+        path = 'file://' + path
+    obj_type = kwargs.pop('type', 'global')
+    if deserialize.__factory and deserialize.__factory[obj_type]:
+        proc_kwargs = kwargs.copy()
+        deserializer = deserialize.__factory[obj_type][0]
+        return deserializer(path, proc_kwargs.pop('vineyard_ipc_socket'), *args, **proc_kwargs)
+    raise RuntimeError('Unable to find a proper IO driver for %s' % path)
```
sighingnow, line 116: Put those two into `vineyard/io/serialization.py`.

sighingnow (approved):
Overall LGTM. Mergable if necessary.

I also want to know comment from @andydiwenzhu.

sighingnow (approved):

Issue #127, opened 2021-01-05
Enhancement: performance tuning for creating metadata, get, persist, and for concurrent connections


sighingnow:
We haven't do benchmark and compare with similar systems, but that is actually necessary.

Conditions that we need to test:

- [ ] allocating blobs: test the memory allocator.
- [ ] creating metadata: test the metadata service
- [ ] persist many objects: test the communication between etcd server.
- [ ] getting metadata: test the performance of metadata lookup
- [ ] getting blobs: test the performance implication of memory mapping
- [ ] per-reserving memory: test the performance of memory allocator

Performance related issues:

+ #179 about putting an arrow table with tons of chunks into vineyard.

sighingnow:
After a batch of related patches now it could be seen as resolved.

PR #205, opened 2021-02-23
A local metaservice that bypass etcd, to check the performance of metadata part.

sighingnow:
## What do these changes do?

Metadata service profiling with gpertools.

## Related issue number

Related to #127, but not fixes it.


sighingnow:
Close as deprecated.

Issue #212, opened 2021-02-25
The interactive between signature, migration, shallow copy and deletion is tricky. 


sighingnow:
**Describe your problem**

+ we have signature for every object.
+ migration doesn't change signature
+ Q: should shallow copy yields a new signature?
+ when deleting objects, its hard to track the dependency of object ids and signatures at the same time.

**Additional context**

None

PR #219, opened 2021-03-03
Reset the signature in deserialization.

sighingnow:
## What do these changes do?

Otherwise there will be an error if the old object being deleted.

Related to #212.

### DIFF
--- a/modules/io/adaptors/deserializer.cc
+++ b/modules/io/adaptors/deserializer.cc
@@ -64,7 +64,9 @@ Status deserialize_helper(
         target.AddMember(kv.key(), sub_target);
       }
     } else {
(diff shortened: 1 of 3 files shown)


siyuan0322 (approved):
LGTM. So we can add the `unload` back in graphscope ci in next PR?

sighingnow:
> So we can add the unload back in graphscope ci in next PR?

You cannot, unless graphscope moves the next release of vineyard :)

siyuan0322:
> > So we can add the unload back in graphscope ci in next PR?
> 
> You cannot, unless graphscope moves the next release of vineyard :)

Of course, that's exactly what I mean 😉

PR #221, opened 2021-03-05
Makes the implementation of GetMallocMapinfo behaves in logarithmic time.

sighingnow:
## What do these changes do?

There's a TODO in the original code referred from plasma.

The patch could be backported to upstream as well.


sighingnow:
Deprecated since the from the implementation detail (we allocate a large mmap file at startup), there won't be many fds. The trivial search is enough.

Issue #280, opened 2021-05-21
Integration with machine learning frameworks


sighingnow:
**Describe your problem**

Part of our summer-2021.

amanchhaparia:
Hello I am interested in solving this issue, Since I am a beginner, I read the docs available in readme, but I cannot understand where to start. Can anybody help with this.

sighingnow:
Hi @amanchhaparia,

Thanks for being interested in project vineyard. For newcomers, you could obtain a sense about how vineyard works in https://v6d.io/notes/getting-started.html and https://v6d.io/notes/divein.html.

Once you have some idea about how vineyard works, you would know why we need strong integration with machine learnings frameworks to make vineyard more useful for the community.

For more background about vineyard, we also have [deck][1], [persentation][2] as well as [a live demo][3].

[1]: https://drive.google.com/file/d/1Bg299lYYhW9Klon3O89VuCXFfFssiqOQ/view?usp=sharing
[2]: https://www.youtube.com/watch?v=p-falphSJq8&list=PLj6h78yzYM2NoiNaLVZxr-ERc1ifKP7n6&t=778
[3]: https://www.youtube.com/watch?v=vPbF1l5nwwQ&list=PLj6h78yzYM2NoiNaLVZxr-ERc1ifKP7n6&t=585

sighingnow:
Hi @andydiwenzhu and @chaitravi-ce,

Thanks for your efforts on the machine learning integration of vineyard. I have reviewed current two WIP pr, but I still think this task requires a careful overall design before merging to main.

#### `DataSet` in machine learning frameworks

+ Tensorflow:

   1. DataSet. from_tensor_slices https://www.tensorflow.org/api_docs/python/tf/data/Dataset
   2. some thing like `tf.TFRecordDataset` https://www.tensorflow.org/api_docs/python/tf/data/TFRecordDataset

+ Pytorch:

   1. TensorDataset https://pytorch.org/docs/stable/data.html, which can be used to wrap one or more tensors.

+ MxNet

    1. NDArrayDataset https://github.com/apache/incubator-mxnet/blob/master/src/io/dataset.cc
    2. ArrayDataset that consumes many `NDArrayDataset`s(columns of a dataframe): https://mxnet.apache.org/api/python/docs/api/gluon/data/index.html#mxnet.gluon.data.ArrayDataset 

**Note that don't mix the concepts and usages of dataset, dataloader and data samplers.**

#### Integration with Vineyard.

[...]

PR #305, opened 2021-06-09
add dataset

andydiwenzhu:
## Related issue number

Fixes

### DIFF
--- a/python/vineyard/data/dataset.py
+++ b/python/vineyard/data/dataset.py
@@ -20,6 +20,7 @@ from vineyard._C import ObjectMeta
 from torch.utils.data import Dataset
 from .utils import from_json, to_json
 
+
 def dataset_builder(client, value, builder):
     meta = ObjectMeta()
     meta['typename'] = 'vineyard::DataSet'
@@ -29,6 +30,7 @@ def dataset_builder(client, value, builder):
         meta.add_member(f'data_{i}_', builder.run(client, data))
         meta.add_member(f'label_{i}_', builder.run(client, label))
     return client.create_metadata(meta)
+
 
 class VineyardDataSet(Dataset):
     '''
@@ -51,8 +53,10 @@ class VineyardDataSet(Dataset):
         label = self.resolver.run(self.obj.member(f'label_{idx}_'))
         return (data, label)
 
+
 def dataset_resolver(obj, resolver):
     return VineyardDataSet(obj, resolver)
+
 
 def register_dataset_types(builder_ctx, resolver_ctx):
     if builder_ctx is not None:
--- a/python/vineyard/data/dataset.py
+++ b/python/vineyard/data/dataset.py
@@ -16,12 +16,6 @@
 # limitations under the License.
 #
 
-import json
-import numpy as np
-import pandas as pd
-from pandas.core.internals.blocks import Block
-from pandas.core.internals.managers import SingleBlockManager
-
 from vineyard._C import ObjectMeta
 from .utils import from_json, to_json
 from torch.utils.data import Dataset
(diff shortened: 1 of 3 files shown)


python/vineyard/data/dataset.py
```diff
@@ -0,0 +1,66 @@
...
+
+def dataset_builder(client, value, builder):
+    meta = ObjectMeta()
+    meta['typename'] = 'vineyard::DataSet'
+    meta['num'] = to_json(len(value))
+    for i in range(len(value)):
+        data, label = value[i]
+        meta.add_member(f'data_{i}_', builder.run(client, data))
+        meta.add_member(f'label_{i}_', builder.run(client, label))
+    return client.create_metadata(meta)
```
sighingnow, line 32: How this dataset interactive with things like `DataFrame` or `Tensor`?

Shouldn't we build `DataSet` from `DataFrame` or `Tensor`, rather than create a new class? How could the new datatype interact with other applications bulid upon vineyard, like pandas as well as GAE?

python/vineyard/data/dataset.py
```diff
@@ -0,0 +1,66 @@
+#! /usr/bin/env python
```
sighingnow, line 1: Put the file to `python/vineyard/contrib` as `python/vineyard/contrib/pytorch.py`, as every ML framework has its own `DataSet`.

python/vineyard/data/dataset.py
```diff
@@ -0,0 +1,66 @@
...
+        self.resolver = resolver
+        self.num = from_json(obj.meta['num'])
+
+    def __len__(self):
+        return self.num
+
+    def __getitem__(self, idx):
+        data = self.resolver.run(self.obj.member(f'data_{idx}_'))
+        label = self.resolver.run(self.obj.member(f'label_{idx}_'))
+        return (data, label)
```
sighingnow, line 54: `data` and `label` usually should have different number of columns?

python/vineyard/data/tests/test_dataset.py
```diff
@@ -0,0 +1,47 @@
...
+class RandomDataset(Dataset):
+    def __init__(self, num):
+        self.num = num
+        self.ds = [(np.random.rand(4, 5, 6), np.random.rand(2, 3)) for i in range(num)]
+
+    def __len__(self):
+        return self.num
+
+    def __getitem__(self, idx):
+        return self.ds[idx]
```
sighingnow, line 37: I'm not very sure, but I think sampling from dataset in pytorch way should be use a `DataLoader`?

sighingnow (approved):

python/vineyard/data/dataset.py
andydiwenzhu: The Dataset type is indeed a composition type from dataframe and tensor. But the structure itself deserves a type.

python/vineyard/data/dataset.py
andydiwenzhu: Nope. I suppose this should be the "common" dataset type we would like to support in vineyard.

python/vineyard/data/dataset.py
andydiwenzhu: Here the shapes of data and label are independent.

python/vineyard/data/dataset.py
sighingnow: But it inherits from pytorch.

Pytorch itself shouldn't be dependent of vineyard.

python/vineyard/data/tests/test_dataset.py
andydiwenzhu: Dataset is the input of DataLoader

python/vineyard/data/dataset.py
sighingnow: The how could map data generated by other engines to `Dataset` and shared with pytorch/tf/mxnet?

sighingnow:
Close as we need reconcile the detailed design in #280.

Issue #200, opened 2021-02-09
Investiage the resource requirement for etcd when deploying vineyard


sighingnow:
**Describe your problem**

During developing GraphScope we found a lot of "execute too long" or "timed out" error if the CPU resource is low (e.g., 2 cores doesn't work however 24 core works pretty well).

We need to investiage the actual requirement, and check if that is a bug/limitation in etcd.

**Additional context**

None

sighingnow:
The etcd docs [says](https://etcd.io/docs/v3.4.0/op-guide/hardware/):

> Few etcd deployments require a lot of CPU capacity. Typical clusters need two to four cores to run smoothly. Heavily loaded etcd deployments, serving thousands of clients or tens of thousands of requests per second, tend to be CPU bound since etcd can serve requests from memory. Such heavy deployments usually need eight to sixteen dedicated cores.

Could @siyuan0322 help to investigate if 4-cores deployment works for your loading graph case ?

PR #313, opened 2021-06-15
Improve the efficiency of vineyardd by avoid establishing new watches in requestUpdates

### LINKED ISSUE #304: Optimize the metadata sync: the `requestUpdate` part: don't trigger a new watcher
**Describe your problem**

This is how vineyard sync metadata from etcd:

1. vineyard has a background watcher
2. every time vineyardd needs to guarantee the metadata is the latest version, vineyard create a new watcher and stop when receive a batch of events.

It is not very performant, as

1. repeated update/processing on metadata tree
2. more importantly, create a watcher is expensive, as it involves at least 4 requests to etcd, and etcd requires a lot of job to create the watcher, and
3. it affects the existing long-running watcher's response time

and it has potential bug, as

1. etcd has a maximum revision size for batch deliever of watched events, i.e., the received events may not be the complete changeset.

That needs to be improved.

**Proposal**

The case should be fixed as

1. keep the long-running watch
2. when vineyard requires the latest version of metadata, it first trigger a `HEAD` request to obtian the latest revision, which should be very cheap, then
3. register a callback into the watcher, and when watcher find expected revisions from a incoming batch, trigger the callback.

Note that the newly created watcher shouldn't be fast/returns early than the long running watcher, according to the implementation of etcd (we need do test, through).

**Additional context**

See also watchable store in etcd for how watch works: https://github.com/etcd-io/etcd/blob/main/server/mvcc/watcher_group.go

andydiwenzhu:
## What do these changes do?

Improve the persist efficiency by avoiding new watches in requestUpdates.
For benchmark, current main branch takes 195.63s for 9 related testcases. While this PR only takes 82.64s (<50%)

## Related issue number

Fixes #304.

### DIFF
--- a/src/server/services/etcd_meta_service.cc
+++ b/src/server/services/etcd_meta_service.cc
@@ -17,6 +17,7 @@ limitations under the License.
 
 #include <chrono>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "boost/asio/steady_timer.hpp"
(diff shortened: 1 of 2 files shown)


sighingnow (changes requested):
Please use a meaningful statement as the title of a pull request.

And verify the pull request with a performance comparison.

sighingnow (approved):
Really nice job!

sighingnow:
I think this issue should resolves #200 as well, as etcd no longer runs under so heavy load.

PR #321, opened 2021-06-19
[FEATURE-316] Condition to check maximum shared memory size

### LINKED ISSUE #316: Flexible parameter to set shared memory limit that vineyard can use.
**Describe your problem**

In the current implementation, we specify a `--size` for maximum shared memory size that vineyard daemon can use. However, it is just a SOFT limit. And,

When there's no enough shared memory (i.e., size of `/dev/shm` path) the limit doesn't make sense. And when users exceed the memory limit, a BUS error will occur at the client side, and crashes the program.

We need to check the maximum shared memory size first, and raise a warning (n.b.: not a failure) to users when it less than the soft limit in the `--size` parameters.

Rohan-cod:

## What do these changes do?
Added a condition to check the **maximum shared memory size** and raise a **warning** when it is **less** than the value of the `size` parameter.

## Related issue number

Fixes #316

### DIFF
--- a/python/vineyard/shared_memory/shared_memory.py
+++ b/python/vineyard/shared_memory/shared_memory.py
@@ -30,8 +30,16 @@ Refer to the documentation of multiprocessing.shared_memory for details.
 import multiprocessing.shared_memory
 import struct
 import warnings
+import os
+import tempfile
 
 from vineyard._C import ObjectID
+
+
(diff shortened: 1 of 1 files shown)


sighingnow:
Such kind of check should be placed in vineyard server (the bulk store module), I believe.

PR #335, opened 2021-06-22
Fixs issues in associate a backtrace to erroneous status.(v6d-io#326)

### LINKED ISSUE #326: Associate a backtrace to erroneous Status
**Describe your problem**

Vineyard returns `Status` when error occurs, and we have things like `VINEYARD_CHECK_OK` to act as assertions to check pre/post conditions in both the client and the server.

However, from the `Status` it is still hard to know what unexpected happens and is not every easy for debugging. We have a set of constructors in `Status` to construct a status. It would be good to associate a backtrace into the Status, at least under debug mode.

We have do similar things in [`RETURN_GS_ERROR`](https://github.com/v6d-io/v6d/blob/main/modules/graph/utils/error.h#L121) and we definitely could do more.

**Additional context**

peilii:
What do these changes do?
-------------------------
Associate a backtrace to erroneous Status.
1. Add a `std::string backtrace_` field into the `class Status`
2. Add a getter `const std::string Status::Backtrace() const {}` to `class Status`
3. In constructor of `Status`, when the status code is not “OK", initialize the `backtrace_` field by calling `vineyard::backtrace_info::backtrace()`.

Related issue number
--------------------
Fixes #326

### DIFF
--- a/src/common/util/status.h
+++ b/src/common/util/status.h
@@ -235,7 +235,6 @@ class VINEYARD_MUST_USE_TYPE Status {
     }
   }
   Status(StatusCode, const std::string& msg);
-  Status(StatusCode, const std::string& msg, const std::string& bt);
   // Copy the specified status.
   inline Status(const Status& s);
   inline Status& operator=(const Status& s);
--- a/src/common/util/status.h
+++ b/src/common/util/status.h
@@ -235,6 +235,7 @@ class VINEYARD_MUST_USE_TYPE Status {
     }
   }
   Status(StatusCode, const std::string& msg);
+  Status(StatusCode, const std::string& msg, const std::string& bt);
   // Copy the specified status.
   inline Status(const Status& s);
   inline Status& operator=(const Status& s);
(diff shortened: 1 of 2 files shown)


sighingnow:
Hi @peilii, please sign-off you commit: https://github.com/v6d-io/v6d/pull/335/checks?check_run_id=2880965191

src/common/util/status.h
```diff
@@ -235,6 +235,7 @@ class VINEYARD_MUST_USE_TYPE Status {
     }
   }
   Status(StatusCode, const std::string& msg);
+  Status(StatusCode, const std::string& msg, const std::string& bt);
```
sighingnow, line 238: That is not need, as we shouldn't expect the user to specify the "backtrace".

sighingnow (approved):
Just some minor comment, otherwise looks good to me.

peilii:
> Hi @peilii, please sign-off your commit: https://github.com/v6d-io/v6d/pull/335/checks?check_run_id=2880965191

I think it is done now.

src/common/util/status.h
peilii: > That is not needed, as we shouldn't expect the user to specify the "backtrace".

Sure. This makes sense.

sighingnow:
Merged, thank you @peilii !

peilii:
> Merged, thank you @peilii !

No problem! Thank you for your guidance!

PR #343, opened 2021-06-28
Initialize the "handled_rev_" with a proper value.

sighingnow:
What do these changes do?
-------------------------

Otherwise when there's no incoming events and we still want to register
a callback on current watcher with HEAD revision (since_rev), the callback
won't be called immediately as expected.

The bug can be reproduced when there's only one instance.

As it is a "daemon watcher", we assume there's only one running backend
watcher.

Related issue number
--------------------

N/A, but related to #313.


andydiwenzhu (approved):
LGTM

Issue #298, opened 2021-06-08
Associate vineyard object with the python value


sighingnow:
**Describe your problem**

For some object types, e.g., `pd.DataFrame`, it is trivial to associate the underlying vineyard object to the python value, however, for some types it is hard due to the limitation of Python, e.g., `np.ndarray`.

Such a feature would help us identify whether an object in Python has a corresponding shared memory value in the vineyard.

The missing of such a feature leads to awkward situations,

```python
x = np.array([1,2,3,4])
x_id = client.put(x)
y = client.get(x_id)
y_id = client.put(y)
```

As the value `y` is obtained from vineyard, and when putting it into vineyard again, vineyard client has no way to figure out that fact and copy the value to vineyard again, which is bad.

**Additional context**

None.

sighingnow:
A possible solition, for every type `T`, generated a type `T'`, which inherits from `T`.

The challenges are:

1. construct an object of `T'` from object of `T`, which should doable by updating the underlying `PyObject*`.
2. efficiency: cache every type that we have processed to avoiding repeatedly creating a type for same original python type.

sighingnow:
> A possible solition, for every type `T`, generated a type `T'`, which inherits from `T`.
> 
> The challenges are:
> 
> 1. construct an object of `T'` from object of `T`, which should doable by updating the underlying `PyObject*`.
> 2. efficiency: cache every type that we have processed to avoiding repeatedly creating a type for same original python type.

That cannot be done in a SAFE way.

sighingnow:
Close as after #346 and #347 it is no longer a severe problem.

PR #347, opened 2021-06-29
Improve the implementation of `create_metadata`.

sighingnow:
What do these changes do?
-------------------------

Returns `metadata` rather than `id` in `client.create_metadata` to avoid  a indirect remote sync/update to bump up performance when creating **nested** python objects to vineyard.

Related issue number
--------------------

Related to #298 and is the subsequential work of #346.

### DIFF
--- a/python/vineyard/_vineyard_docs.py
+++ b/python/vineyard/_vineyard_docs.py
@@ -284,8 +284,7 @@ Parameters:
         The name of the member object.
     object: :class:`Object`, :class:`ObjectID` or :class:`ObjectMeta`
         The reference to the member object or the object id of the member object.
-'''
-)
+''')
 
 add_doc(
     ObjectID, r'''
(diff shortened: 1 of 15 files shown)


python/vineyard/_vineyard_docs.py
```diff
@@ -394,28 +411,53 @@ def add_doc(target, doc):
         Objects that will be deleted. The :code:`object_id` can be a single :class:`ObjectID`, or a list of
         :class:`ObjectID`.
     force: bool
-        Forcely delete an object means the member will be recursively deleted even if the
-        member object is also refered by others. The default value is :code:`True`.
+        Forcedly delete an object means the member will be recursively deleted even if the
+        member object is also refereed by others. The default value is :code:`True`.
```
siyuan0322, line 415: referred?

siyuan0322 (approved):

python/vineyard/_vineyard_docs.py
sighingnow: Fixed, thanks!

PR #350, opened 2021-06-30
Graceful clean-up on bulk store.(v6d-io#333)

### LINKED ISSUE #333: Graceful clean-up on bulk store: "free" the memory when exit.
Describe your problem
---------------------

Vineyard server (`vineyardd`) holds all memory blocks allocated by the client. At exit there are still memory chunks that are not freed.

If vineyardd itself is gracefully exits (receiving a `CTRL-C` signal), it would be better to free those shared memory segments as well.

If it is a **feature request**, please provides a clear and concise description of what you want to happen:

1. What is the problem:

   Shared memory is not freed when exit.

2. The behavior that you expect to work:

   There should be no "memory leaks", and can be checked via a memory sanitizer / address sanitizer, see also #302. 

Additional context
------------------

peilii:
Signed-off-by: Pei Li <peili@andrew.cmu.edu>

What do these changes do?
-------------------------
Free objects in BulkStore.

Related issue number
--------------------

Fixes #333

### DIFF
--- a/src/server/memory/memory.cc
+++ b/src/server/memory/memory.cc
@@ -128,6 +128,13 @@ static void recycle_arena(const uintptr_t base, const size_t size,
 
 std::set<ObjectID> BulkStore::Arena::spans{};
 
+BulkStore::~BulkStore() {
+  for (auto iter = objects_.begin(); iter != objects_.end(); iter++) {
+    auto& object = iter->first;
+    VINEYARD_DISCARD(Delete(object));
+  }
+}
+
 Status BulkStore::PreAllocate(const size_t size) {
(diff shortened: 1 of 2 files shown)


peilii:
@sighingnow Hi. I think this branch compiles in my Mac, however it fails to build with the following error message while installing dependencies:
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?

Would you please offer me some advice about what could possibly go wrong?

sighingnow:
Thank you @peilii, this pull request looks good to me, and I will take care of the CI failure.

sighingnow (approved):

PR #351, opened 2021-06-30
Fixes the buffer overflow error in get_demangled_name.

sighingnow:
What do these changes do?
-------------------------

The `abi::__cxa_demangle()` may fail, and returns a `NULL`, but there indeed a buffer that can be reused and the length is indeed a non-zero value.

Related issue number
--------------------

Fixes alibaba/GraphScope#373


siyuan0322 (approved):
Nice work 👍

sighingnow:
> Nice work 👍

Thank you for the help @siyuan0322 !

PR #360, opened 2021-07-02
Fixes the bug in requestUpdates, and add a check for the invariant.

sighingnow:
What do these changes do?
-------------------------

Fixes the expected revision in `EtcdMetaService::requestUpdates()`, and add a concurrent persist+sync test case.

Related issue number
--------------------

Follow-up work of #313.


andydiwenzhu (approved):
LGTM

PR #368, opened 2021-07-08
Do not ignore the return value of `realpath`

siyuan0322:


modules/io/io/io_factory.cc
```diff
@@ -61,7 +61,11 @@ std::unique_ptr<IIOAdaptor> IOFactory::CreateIOAdaptor(
     if (!s.ok()) {  // Assume it's a local file
       // defaulting to local file system: resolve to abs path first
       char resolved_path[PATH_MAX];
-      realpath(location_to_parse.c_str(), resolved_path);
+      char* res = realpath(location_to_parse.c_str(), resolved_path);
+      if (!res) {
+        LOG(ERROR) << "Failed to resolve realpath of " << location_to_parse;
+        return nullptr;
+      }
```
sighingnow, line 68: It is expected. When `location_to_parse` doesn't exist, it returns failure but `resolved_path` will be still filled as the expanded path.

The open mode might be `w` and the file doesn't exist, it still returns fail, but it is valid.

Ignore the return value of `realpath` is OK here and we just delegates the failure reporting the the IO adaptor itself.

sighingnow (changes requested):

modules/io/io/io_factory.cc
siyuan0322: Then I changed it to log an warning. The ignoring behavior is treated as an warning in gcc-9, and with -Werror enabled it's an error.

sighingnow: If you want to check if it resolves successfully, you need to check if `resolved_path` is empty after resolve. It it is empty, you could raise a the warning.

sighingnow (changes requested):

modules/io/io/io_factory.cc
siyuan0322: If it failed the return value will be NULL.

sighingnow: For `w`, it fails but `resolved_path` will be filled successfully. The we SHOULDN'T raise this warning.

Instead, you need to check if `resolved_path` is null (is empty string).

siyuan0322: I don't want to check resolved_path... The point is that I want to make use of the return value, you can treat the VLOG(2) as a debug message anyway :)

sighingnow (approved):

PR #394, opened 2021-07-29
Add DMatrix resolvers and tests for better integration with xgboost 

andydiwenzhu:

What do these changes do?
-------------------------

As title, we omit the integration with global objects in this PR since xgboost does not support native distributed training.

Instead, we will try to handle global objects in distributed training frameworks that xgboost leverages (e.g., dask) in the future.

Related issue number
--------------------

Related to #280.

### DIFF
--- a/python/vineyard/contrib/ml/tests/test_xgboost.py
+++ b/python/vineyard/contrib/ml/tests/test_xgboost.py
@@ -1,0 +1,81 @@
+#! /usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020-2021 Alibaba Group Holding Limited.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import numpy as np
+import pandas as pd
+
+import pyarrow as pa
+import pytest
+
+import vineyard
+from vineyard.core import default_builder_context, default_resolver_context
+from vineyard.contrib.ml.xgboost import register_xgb_types
+
+register_xgb_types(default_builder_context, default_resolver_context)
+
+
+def test_numpy_ndarray(vineyard_client):
+    arr = np.random.rand(4, 5)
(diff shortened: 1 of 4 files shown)


python/vineyard/contrib/ml/tests/test_xgboost.py
```diff
@@ -0,0 +1,87 @@
...
+#
+
+import numpy as np
+import pandas as pd
+
+try:
+    import scipy as sp
+    import scipy.sparse
+except ImportError:
+    sp = None
```
sighingnow, line 26: `ps` is not used.

python/vineyard/contrib/ml/xgboost.py
```diff
@@ -0,0 +1,114 @@
...
+        blocks.append(Block(values, placement, ndim=2))
+    index = np.arange(index_size)
+    df = pd.DataFrame(BlockManager(blocks, [pd.Index(columns), index]))
+    if 'label' in kw:
+        label = df.pop(kw['label'])
+        # data column can only be specified if label column is specified
+        if 'data' in kw:
+            df = np.stack(df[kw['data']].values)
+        return xgb.DMatrix(df, label)
+    return xgb.DMatrix(df)
```
sighingnow, line 78: Here could we explicitly invoke the `dataframe_resolver` to get a `pd.DataFrame` instead? To avoid copying the copy the code again and again.

You could revisit the resolver for tensorflow dataset as well.

test/runner.py
```diff
@@ -412,6 +427,10 @@ def main():
         with start_etcd() as (_, etcd_endpoints):
             run_io_adaptor_distributed_tests(etcd_endpoints, args.with_migration)
 
+    if args.with_contrib:
+        with start_etcd() as (_, etcd_endpoints):
+            run_python_contrib_tests(etcd_endpoints)
```
sighingnow, line 432: Put the `run_python_contrib_tests(etcd_endpoints)` under `if args.with_python`.

test/runner.py
```diff
@@ -385,13 +398,15 @@ def parse_sys_args():
                             help='Whether to run IO adaptors tests')
     arg_parser.add_argument('--with-migration', action='store_true', default=False,
                             help='Whether to run object migration tests')
+    arg_parser.add_argument('--with-contrib', action='store_true', default=False,
```
sighingnow, line 401: `--with-contrib-ml` or `--with-ml` would be better.

sighingnow (changes requested):

python/vineyard/contrib/ml/xgboost.py
andydiwenzhu: We will leave it here and update builder/resolver in the future as discussed

sighingnow (approved):

sighingnow:
Landed. Thanks for the effort!

PR #395, opened 2021-07-29
Added Global Resolvers for Tensorflow and Local Builders and Resolvers for PyTorch

chaitravi-ce:

What do these changes do?
-------------------------
Added resolvers for vineyard::GlobalTensor and vineyard::GlobalDataFrame which will return a list of local objects in form of tf.data.Datasets.

Related issue number
--------------------

Related to #280

### DIFF
--- a/python/vineyard/contrib/ml/tensorflow.py
+++ b/python/vineyard/contrib/ml/tensorflow.py
@@ -30,6 +30,7 @@ except:
 from pandas.core.internals.managers import BlockManager
 import numpy as np
 import tensorflow as tf
+import vineyard
 
 
 def tf_tensor_builder(client, value, **kw):
@@ -82,6 +83,16 @@ def tf_builder(client, value, builder, **kw):
             return tf_dataframe_builder(client, value, builder, **kw)
         else:
             return tf_tensor_builder(client, value, **kw)
+
+
+def tf_create_global_tensor(client, value, builder, **kw):
+    # TODO
+    pass
+
+
+def tf_create_global_dataframe(client, value, builder, **kw):
+    # TODO
+    pass
 
 
 def tf_tensor_resolver(obj):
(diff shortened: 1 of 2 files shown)


python/vineyard/contrib/ml/tensorflow.py
```diff
@@ -145,6 +156,34 @@ def tf_table_resolver(obj, resolver):
...
+    endpoint = kw.get('endpoint')
+    client = vineyard.connect(endpoint)
+    num = from_json(meta['num'])
+    partition_index = from_json(meta['partition_index_'])
+    data = []
+    for i in range(partition_index):
+        loc = meta[f'partition_{i}']
+        if loc.islocal:
+            data.append(client.get(loc.id))
+    return data
```
andydiwenzhu, line 184: Could you check the from_generator method (https://www.tensorflow.org/api_docs/python/tf/data/Dataset#from_generator) and return a single dataset instead of a list of datasets. So as the case in global tensor resolver.

andydiwenzhu (commented):
Please refer to the test_xgboost and add tests for your tensorflow integrations.
I will add support for global objects on xgboost soon, you can add tests for local objects first.

python/vineyard/contrib/ml/tensorflow.py
chaitravi-ce: Used tf.concatenate , it will also return a single dataset.

python/vineyard/contrib/ml/tensorflow.py
```diff
@@ -145,6 +156,40 @@ def tf_table_resolver(obj, resolver):
...
+    meta = obj.meta
+    endpoint = kw.get('endpoint')
+    client = vineyard.connect(endpoint)
+    num = from_json(meta['num'])
+    partition_index = from_json(meta['partition_index_'])
+    data = []
+    for i in range(partition_index):
+        loc = meta[f'partition_{i}']
+        if loc.islocal:
+            data.append(client.get(loc.id))
```
andydiwenzhu, line 169: Resolving a partition in a global object is similar to resolving a tensor in a dataframe. Please refer to (https://github.com/v6d-io/v6d/blob/44bf8f47fca10f2e7559f0865d7285a7a2c49ff4/python/vineyard/data/dataframe.py#L58)
to avoid establishing another client and request to the server which is a waste of resources.

andydiwenzhu (approved):
LGTM

PR #399, opened 2021-07-30
Contextual resolver/builder/driver for flexible nested dynamic registry.

### LINKED ISSUE #398: Stackable {resolver,builder}_context for flexible builder/resolver selection.
Describe your problem
---------------------

There are cases where we need to mixuse of a set of builder/resolver in the same context, e.g., when resolving a `pd.DataFrame` as a `xgboost` dataset ( see also #394 ), we need to

+ resolve `numpy.ndarray` in the index value into a numpy array
+ resolve `numpy.ndarray` in the columns payload into a xgboost `DMatrix`

Our current implementation for the builder/resolve yields a "overwrite" behaviour, which doesn't match with our expectation when resolving/building a complex data structure.

Additional context
------------------

The resolver/builder context needs to be

+ nested and stackable
+ support fast lookup for certain typename and certain python type
+ provide flexible syntax

sighingnow:
What do these changes do?
-------------------------

Implemented context as described in #398.

Related issue number
--------------------

Fixes #398.

### DIFF
--- a/python/vineyard/__init__.py
+++ b/python/vineyard/__init__.py
@@ -98,7 +98,8 @@ from . import io
 from . import launcher
 from . import shared_memory
 
-from .core import default_builder_context, default_resolver_context, default_driver_context
+from .core import default_builder_context, default_resolver_context, default_driver_context, \
+    builder_context, resolver_context, driver_context
 from .data import register_builtin_types
 from .data.graph import Graph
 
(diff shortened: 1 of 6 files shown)


andydiwenzhu (commented):
Can merge after CI. Thanks for the quick fix!

PR #401, opened 2021-07-31
Adds new commands, arguments and examples to vineyard-ctl

Rohan-cod:

What do these changes do?
-------------------------

- Added `examples` for all the **commands** in **vineyard-ctl**.

- Added `copy` and `head` **commands** to **vineyard-ctl**.

- Added a new argument `regex_pattern` to the `del` **command**. It allows to **delete** all the **objects** that match the **regex pattern**.

- Added more **configuration options** to the `put` **command**.


python/vineyard/cli.py
```diff
@@ -29,10 +29,23 @@
...
+                '--ipc_socket /var/run/vineyard.sock\n\n2. List vineyard objects\n' +
+                '>>> vineyard-ctl ls --limit 6\n\n3. Query a vineyard object\n>>> ' +
+                'vineyard-ctl query --object_id 00002ec13bc81226\n\n4. Print first n lines' +
+                ' of a vineyard object\n>>> vineyard-ctl head --object_id 00002ec13bc81226' +
+                ' --limit 3\n\n5. Copy a vineyard object\n>>> vineyard-ctl copy --object_id' +
+                ' 00002ec13bc81226 --shallow\n\n6. Delete a vineyard object\n>>> vineyard-ctl' +
+                ' del --object_id 00002ec13bc81226\n\n7. Get the status of connected vineyard' +
+                ' server\n>>> vineyard-ctl stat\n\n8. Put a python value to vineyard\n>>> ' +
+                'vineyard-ctl put --file example_csv_file.csv\n\n9. Edit configuration file\n' +
+                '>>> vineyard-ctl config --ipc_socket_value /var/run/vineyard.sock'))
```
sighingnow, line 48: Could you please put those lines as a miltiple-line string like

```python
examples = """
Some examples on how to use vineyard-ctl:

1. Connect to a vineyard server

    >>> vineyard-ctl --ipc_socket /var/run/vineyard.sock

2. List vineyard objects

    ....
"""
```

and use it like

```python
epilog = example,
```

siyuan0322: Yes, multiple-line string could preserve the original formats and spaces, make your life easier XD

Rohan-cod: I just updated the examples to be a multi-line string. :)

python/vineyard/cli.py
```diff
@@ -254,14 +340,36 @@ def put_object(client, args):
...
             raise Exception(('The following error was encountered while putting ' +
                              f'{args.file} as pandas dataframe to vineyard server:\n{exc}')) from exc
 
 
+def head(client, args):
+    """Utility to print the first n lines of a vineyard object."""
+    try:
+        value = client.get(as_object_id(args.object_id))
+        if isinstance(value, pd.DataFrame):
+            print(value.head(args.limit))
```
siyuan0322, line 356: else:
  print('Only support head a dataframe yet')

Or better you could fetch the meta first and check the type.

python/vineyard/cli.py
```diff
@@ -221,14 +292,29 @@ def query(client, args):
...
+            objects = client.list_objects(pattern=args.regex_pattern, regex=True, limit=5)
+        except BaseException as exc:
+            raise Exception(f'The following error was encountered while listing vineyard objects:\n{exc}') from exc
+        for obj in objects:
+            try:
+                client.delete(object_id=as_object_id(obj.id), force=args.force, deep=args.deep)
+                print(f'The vineyard object({obj.id}) was deleted successfully')
+            except BaseException as exc:
+                raise Exception(('The following error was encountered while deleting ' +
+                                 f'the vineyard object({obj.id}):\n{exc}')) from exc
```
siyuan0322, line 315: If you do `from exc`, then you don't need to print to include the `exc` in the Exception's body. Python will print the trace.

siyuan0322: You could simply do ` raise Exception(f'Error occur when deleting the vineyard object ({obj.id})') from exc`.
Same for the others.

siyuan0322 (approved):

PR #403, opened 2021-08-02
Run xgboost tests use a scoped fixture, rather than populating the global ctx

sighingnow:
What do these changes do?
-------------------------

Use a scoped autouse fixture to run xgboost tests.

Related issue number
--------------------

Follow-up work for #394 and #399.

### DIFF
--- a/python/vineyard/contrib/ml/tests/test_xgboost.py
+++ b/python/vineyard/contrib/ml/tests/test_xgboost.py
@@ -24,9 +24,17 @@ import pytest
 
 import vineyard
 from vineyard.core import default_builder_context, default_resolver_context
+from vineyard.core.builder import builder_context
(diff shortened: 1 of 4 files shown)


python/vineyard/contrib/ml/tests/test_xgboost.py
```diff
@@ -24,9 +24,17 @@
 
 import vineyard
 from vineyard.core import default_builder_context, default_resolver_context
```
andydiwenzhu, line 26: This line can be removed?

andydiwenzhu (commented):
LGTM

python/vineyard/contrib/ml/tests/test_xgboost.py
sighingnow: Fixed. thanks!

PR #409, opened 2021-08-04
Add Dask Integration: vineyard as the data source for dask

andydiwenzhu:

What do these changes do?
-------------------------

Add two resolvers for dask: GlobalTensor --> dask.Array, GlobalDataFrame --> dask.dataframe

Related issue number
--------------------

Part of #412.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -274,6 +274,10 @@ jobs:
           rm -rf build/lib.* build/bdist.*
           python3 setup_ml.py bdist_wheel
 
+          # build & install vineyard-dask
+          rm -rf build/lib.* build/bdist.*
+          python3 setup_dask.py bdist_wheel
+
           # build & install vineyard-io
           pushd modules/io
           rm -rf build/lib.* build/bdist.*
@@ -325,7 +329,7 @@ jobs:
 
             rm -rf default.etcd
             rm -rf /dev/shm/etcd*
-            python3 test/runner.py --with-python --with-migration --with-contrib-ml
+            python3 test/runner.py --with-python --with-migration --with-contrib
 
       - name: Run Python Tests
         if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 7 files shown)


python/vineyard/contrib/ml/distributed/dask.py
```diff
@@ -0,0 +1,98 @@
...
+def dask_dataframe_builder(client, value, builder, **kw):
+    # TODO: build dask.dataframe to vineyard::GlobalDataFrame
+    pass
+
+
+def dask_array_resolver(obj, resolver, **kw):
+    def get_partition(socket, obj_id):
+        client = vineyard.connect(socket)
+        np_value = client.get(obj_id)
+        return da.from_array(np_value)
```
sighingnow, line 43: Is there a copy? (I'm not sure.

sighingnow (commented):
Could we have a package called vineyard-dask? under vineyard/contrib/dask.

It is quite strange to put it under ml/distributed....

sighingnow:
Seems the resolver assumes both the tensor and dataframe are chunked along the `x` ax,  it not alway be true, as the tensor/dataframe may come from mars.

We haven't put enough efforts in other ML integrations, I know it is not trivial, but it does need to be fixed. As the dask already natively supports split chunks along both the `x` and `y` axis, and we have already record enough information in the metadata, could we do right thing here for dask? Otherwise it would be depreacated and replaced by a new implementation someday.

andydiwenzhu:
> Seems the resolver assumes both the tensor and dataframe are chunked along the `x` ax, it not alway be true, as the tensor/dataframe may come from mars.
> 
> We haven't put enough efforts in other ML integrations, I know it is not trivial, but it does need to be fixed. As the dask already natively supports split chunks along both the `x` and `y` axis, and we have already record enough information in the metadata, could we do right thing here for dask? Otherwise it would be depreacated and replaced by a new implementation someday.

sighingnow:
Andy @andydiwenzhu  I didn't mean closing the issue. We could

1. put those things into a vineyard-dask package
2. leave a check to raise an `NotImplementedError` exception when the source is not chunked as what we want

Then we could get this pull request landed.

andydiwenzhu:
> Andy @andydiwenzhu I didn't mean closing the issue. We could
> 
> 1. put those things into a vineyard-dask package
> 2. leave a check to raise an `NotImplementedError` exception when the source is not chunked as what we want
> 
> Then we could get this pull request landed.

OK. That was a bad click, don't worry :)

sighingnow (approved):

Issue #412, opened 2021-08-05
Integration vineyard with dask


sighingnow:
Describe your problem
---------------------

- [ ] As the data source/sink
   - #409
- [ ] As the underlying in-memory storage for `dask.persist`

Additional context
------------------

Part of #372.

sighingnow:
Reopen as part is still unresolved.

PR #415, opened 2021-08-05
Add Dask Integration: vineyard as data sink

andydiwenzhu:

What do these changes do?
-------------------------

Add builders for dask:

- dask.Array --> vineyard::GlobalTensor
- dask.DataFrame --> vineyard::GlobalDataFrame

Related issue number
--------------------

Part of #412.

### DIFF
--- a/python/vineyard/contrib/dask/dask.py
+++ b/python/vineyard/contrib/dask/dask.py
@@ -25,21 +25,58 @@ from vineyard._C import ObjectMeta
 
 import dask.array as da
 import dask.dataframe as dd
+import numpy as np
+import pandas as pd
 
 
 def dask_array_builder(client, value, builder, **kw):
-    # TODO: build dask.array to vineyard::GlobalTensor
-    pass
+    def put_partition(v, block_id=None):
+        client = vineyard.connect()
+        obj_id = client.put(v, partition_index=block_id)
+        client.persist(obj_id)
+        return np.array([[int(obj_id)]])
+
+    _ = Client(kw['dask_scheduler'])  #enforce distributed scheduling
+    ids = value.map_blocks(put_partition, dtype=int).compute().flatten()
+    meta = ObjectMeta()
+    meta['typename'] = 'vineyard::GlobalTensor'
+    meta.set_global(True)
+    meta['partitions_-size'] = len(ids)
+
+    for i, obj_id in enumerate(ids):
+        meta.add_member('partitions_-%d' % i, vineyard.ObjectID(obj_id))
+
+    gtensor_meta = client.create_metadata(meta)
+    client.persist(gtensor_meta)
+    return gtensor_meta
 
 
 def dask_dataframe_builder(client, value, builder, **kw):
-    # TODO: build dask.dataframe to vineyard::GlobalDataFrame
-    pass
+    def put_partition(v, partition_info=None):
+        client = vineyard.connect()
+        obj_id = client.put(v, partition_index=(partition_info['number'], 0))
+        client.persist(obj_id)
+        return pd.DataFrame([{'no': partition_info['number'], 'id': int(obj_id)}])
+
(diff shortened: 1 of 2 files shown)


sighingnow (changes requested):
Could you please add some roundtrip test? I mean, putting a dask tensor/dataframe to vineyard, and get it back, then assert on the equality.

python/vineyard/contrib/dask/dask.py
```diff
@@ -25,21 +25,58 @@
...
 def dask_array_builder(client, value, builder, **kw):
-    # TODO: build dask.array to vineyard::GlobalTensor
-    pass
+    def put_partition(v, block_id=None):
+        client = vineyard.connect()
+        obj_id = client.put(v, partition_index=block_id)
+        client.persist(obj_id)
+        return np.array([[int(obj_id)]])
+
+    _ = Client(kw['dask_scheduler'])
```
sighingnow, line 39: Leave a note here to explain why we need such an explicit unused client.

python/vineyard/contrib/dask/dask.py
```diff
@@ -25,21 +25,58 @@
...
+    meta['typename'] = 'vineyard::GlobalTensor'
+    meta.set_global(True)
+    meta['partition_-size'] = len(ids)
+
+    for i, obj_id in enumerate(ids):
+        meta.add_member('partition_-%d' % i, vineyard.ObjectID(obj_id))
+
+    gtensor = client.create_metadata(meta)
+    client.persist(gtensor)
+    return gtensor
```
sighingnow, line 51: `gtensor` -> `gtensor_id` as it is an object id, not a tensor.

python/vineyard/contrib/dask/dask.py
```diff
@@ -25,21 +25,58 @@
...
+    meta['typename'] = 'vineyard::GlobalDataFrame'
+    meta.set_global(True)
+    meta['partition_-size'] = len(res)
+
+    for i in range(len(res)):
+        meta.add_member('partition_-%d' % i, vineyard.ObjectID(res.loc[i]))
+
+    gdf = client.create_metadata(meta)
+    client.persist(gdf)
+    return gdf
```
sighingnow, line 74: `gdf` -> `gdf_id`.

sighingnow: to avoid possible misunderstandling.

python/vineyard/contrib/dask/dask.py
```diff
@@ -25,21 +25,58 @@
...
 
 
 def dask_dataframe_builder(client, value, builder, **kw):
-    # TODO: build dask.dataframe to vineyard::GlobalDataFrame
-    pass
+    def put_partition(v, partition_info=None):
+        client = vineyard.connect()
+        obj_id = client.put(v, partition_index=(partition_info['number'], 0))
+        client.persist(obj_id)
+        return pd.DataFrame([{'no': partition_info['number'], 'id': int(obj_id)}])
```
sighingnow, line 59: Could we return a `vineyard.ObjectID` directly? Why we need to wrap it into a dataframe?

python/vineyard/contrib/dask/dask.py
```diff
@@ -25,21 +25,58 @@
...
 
 
 def dask_array_builder(client, value, builder, **kw):
-    # TODO: build dask.array to vineyard::GlobalTensor
-    pass
+    def put_partition(v, block_id=None):
+        client = vineyard.connect()
+        obj_id = client.put(v, partition_index=block_id)
+        client.persist(obj_id)
+        return np.array([[int(obj_id)]])
```
sighingnow, line 37: We we need to wrap it as an array?

sighingnow: Just image something like `ndarray.sum`, or something like that.

Issue #275, opened 2021-05-13
Vineyard + Prometheus: made monitoring easy


sighingnow:
**Describe your problem**

Vineyard should be visible for prometheus on Kubernetes.

**Additional context**

N/A.

PR #418, opened 2021-08-07
[MRG]🎉 init prometheus setting

univerone:
What do these changes do?
-------------------------

- add basic metrics to log: data_request_latency and data_requests_total

image

-  add [grok_exporter](https://github.com/fstab/grok_exporter) setting file to export those metrics from log files, metrics will be displayed on : <http://localhost:9144/metrics>

```bash
./grok_exporter -config v6d/prometheus/grok_exporter.yml -disable-exporter-metrics
```
image

-  add [prometheus](https://prometheus.io/) setting file to read those metrics

```bash
./prometheus --config.file=v6d/prometheus/prometheus.yml
```
-  add a prometheus recording rule file to calculate average request latency/

image

Related issue number #275 
--------------------

Fixes #issue number

### DIFF
--- a/src/server/server/vineyard_server.cc
+++ b/src/server/server/vineyard_server.cc
@@ -343,7 +343,7 @@ Status VineyardServer::CreateData(
       },
       boost::bind(callback, _1, id, signature, _2));
   double endTime = GetCurrentTime();
-  LOG_SUMMARY("data_request_latency", "create",  endTime - startTime);
+  LOG_SUMMARY("data_request_latency", "create", endTime - startTime);
   LOG_COUNTER("data_requests_total", "create");
   return Status::OK();
 }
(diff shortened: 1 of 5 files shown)


src/server/server/vineyard_server.cc
```diff
@@ -261,6 +262,9 @@ Status VineyardServer::GetData(const std::vector<ObjectID>& ids,
           return status;
         }
       });
+  double endTime = GetCurrentTime();
+  LOG_SUMMARY("data_request_latency", "get", endTime - startTime);
+  LOG_COUNTER("data_requests_total", "get");
```
sighingnow, line 267: The handlers in socket server runs in an async manner, thus the end time should be moved before the `return status;` in the callback.

sighingnow:
Merged, thanks for contributing to vineyard!

PR #424, opened 2021-08-10
Fix template type in VertexDataColum

### LINKED ISSUE #426: The template parameter in `vertex_data_column` is not correct
Describe your problem
---------------------

https://github.com/v6d-io/v6d/blob/38d695796d1a6bb65e5559746763fe8030aabc84/modules/graph/fragment/arrow_fragment.h#L372-L382

The template parameter for `VertexDataColumn` should be `VID_T`, rather than `vertex_t`. The code is not used in both vineyard and graphscope, but it is wrong.

Additional context
------------------

zhanglei1949:
What do these changes do?
-------------------------
Fix the type parameters inconsistency in arrow_fragment::vertex_data_column

Related issue number
--------------------

Fixes #426

### DIFF
--- a/modules/graph/fragment/arrow_fragment.h
+++ b/modules/graph/fragment/arrow_fragment.h
@@ -370,13 +370,13 @@ class ArrowFragment
   }
 
   template <typename DATA_T>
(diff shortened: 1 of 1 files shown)


sighingnow (approved):
Nice catch!

LGTM.

PR #425, opened 2021-08-10
Refine the resolver context implementation, and refactor duplicated code

### LINKED ISSUE #423: Resolver in Python: behaviour of binding vineyard object to resolved result is wrong for nested types.
Describe your problem
---------------------

When "client.get()" object from vineyard, we will associated the vineyard object to the resolved value as the `__vineyard_ref` property to setup such kind of connections. But for nested test, e.g., `pd.DataFrame`, we only assoicate for the outer-most value, which is wrong.

Additional context
------------------

sighingnow:
What do these changes do?
-------------------------

This pull request fixes #423, by patching `__vineyard_ref` in the resolver, and add a `__call__` method for builder/resolver/driver context.

This pull request also reduces the code duplication in the ML integration to make the code more concise.

Related issue number
--------------------

Fixes #423.

### DIFF
--- a/python/vineyard/contrib/ml/pytorch.py
+++ b/python/vineyard/contrib/ml/pytorch.py
@@ -16,21 +16,13 @@
 # limitations under the License.
 #
 
-from vineyard._C import ObjectMeta
-from vineyard.data.utils import from_json, to_json, build_numpy_buffer, normalize_dtype
-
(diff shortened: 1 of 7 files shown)


sighingnow:
@andydiwenzhu Andy, any comments on this pull request?

PR #452, opened 2021-08-18
Added MXNet Builders, Resolvers and Tests

chaitravi-ce:

What do these changes do?
-------------------------

* Added Local Builders and Resolvers for MxNet
* Added Global Resolvers for MxNet
* Added Tests for Local Object

Related issue number
--------------------

Related to #280

### DIFF
--- a/python/vineyard/contrib/ml/mxnet.py
+++ b/python/vineyard/contrib/ml/mxnet.py
@@ -140,7 +140,7 @@ def mxnet_global_dataframe_resolver(obj, resolver, **kw):
     return mx.gluon.data.ArrayDataset((data, label))
 
 
-def register_tensor_types(builder_ctx, resolver_ctx):
+def register_mxnet_types(builder_ctx, resolver_ctx):
     if builder_ctx is not None:
         builder_ctx.register(mx.gluon.data.ArrayDataset, mxnet_builder)
 
--- a/python/vineyard/contrib/ml/mxnet.py
+++ b/python/vineyard/contrib/ml/mxnet.py
@@ -20,9 +20,8 @@ import mxnet as mx
 
 import numpy as np
 import pandas as pd
-from vineyard.core import default_builder_context, default_resolver_context
+from vineyard.core.resolver import resolver_context, default_resolver_context
 
-import vineyard
 from vineyard._C import ObjectMeta
 from vineyard.data.utils import from_json, to_json, build_numpy_buffer, normalize_dtype
 
@@ -99,8 +98,8 @@ def mxnet_record_batch_resolver(obj, **kw):
         records = resolver(obj, **kw)
     records = records.to_pandas()
     if 'label' in kw:
-        target = df[kw['label']].values.astype(np.float32)
-        data = df.drop(kw['label'], axis=1).values.astype(np.float32)
+        target = records[kw['label']].values.astype(np.float32)
+        data = records.drop(kw['label'], axis=1).values.astype(np.float32)
         return mx.gluon.data.ArrayDataset((data, target))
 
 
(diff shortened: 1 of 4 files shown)


python/vineyard/contrib/ml/mxnet.py
```diff
@@ -17,3 +17,163 @@
...
+        else:
+            placement = slice(idx, idx + 1, 1)
+        values = np.expand_dims(np_value, 0)
+        blocks.append(Block(values, placement, ndim=2))
+    if 'index_' in meta:
+        index = resolver.run(obj.member('index_'))
+    else:
+        index = pd.RangeIndex(index_size)
+    df = pd.DataFrame(BlockManager(blocks, [pd.Index(columns), index]))
+    target = df[label].values.astype(np.float32)
```
sighingnow, line 118: Could you please refer the resolvers in `tensorflow.py` for how to reuse `resolver_context()` to avoid such code duplication?

sighingnow (commented):
Overall LGTM.

python/vineyard/contrib/ml/mxnet.py
chaitravi-ce: Yes, I tried that but it gives reshape error.

andydiwenzhu: Is there any difference between the pandas dataframe you get here and the one with resolver_context()? If not, I didn't any reason why that dataframe will have a reshape error?

chaitravi-ce: There is a slight difference. Instead of directly using resolver.run, I have changed it to returning memoryview of the member "buffer"

andydiwenzhu: I don't think so. Inside resolver.run, it also uses np.from_buffer(memoryview(obj.member('buffer_')).

chaitravi-ce: Yes but here there is no reshaping

andydiwenzhu: But why the reshape can be removed? Without the reshape, the np_value will have a different shape than it is saved to vineyard. Then you still want to make a dataframe from this np_value? I don't think this will work. Correct me if I'm wrong.

andydiwenzhu: You post the error log when using resolver_context() here or on Slack. We can help check the log.

chaitravi-ce: Yes, Reshaping is important.

python/vineyard/data/utils.py
```diff
@@ -41,7 +41,7 @@ def normalize_dtype(dtype, dtype_meta=None):
     if dtype in ['u64', 'uint64', 'uint64_t']:
         return np.dtype('uint64')
     if dtype in ['float', 'float32']:
-        return np.dtype('float')
+        return np.dtype('float32')
```
andydiwenzhu, line 44: I think this should be if dtype == 'float': return np.dtype('float')  if dtype == 'float32': return np.dtype('float32') ?

chaitravi-ce: Will Change the code accordingly

python/vineyard/contrib/ml/mxnet.py
```diff
@@ -17,3 +17,142 @@
...
+
+from vineyard._C import ObjectMeta
+from vineyard.data.utils import from_json, to_json, build_numpy_buffer, normalize_dtype
+
+from pandas.core.internals.managers import BlockManager
+try:
+    from pandas.core.internals.blocks import BlockPlacement, NumpyBlock as Block
+except:
+    BlockPlacement = None
+    from pandas.core.internals.blocks import Block
```
sighingnow, line 33: Drop those unused imports please.

sighingnow (approved):

PR #454, opened 2021-08-19
Add version number macro in headers.

sighingnow:
What do these changes do?
-------------------------

Thus we can programmably differ the interface changes between versions.

Related issue number
--------------------

N/A.


sighingnow:
Close as already presented in main.

PR #466, opened 2021-08-27
Update __init__.py to import VineyardXCom

andtheysay:
When trying to use the XCom provider Airflow cannot find the VineyardXCom class because it is not defined in the \_\_init\_\_.py file so it needs to be defined.

What do these changes do?
-------------------------

Fix error when trying to use the XCom backend.

Related issue number
--------------------
None

### DIFF
--- a/python/vineyard/contrib/airflow/xcom/__init__.py
+++ b/python/vineyard/contrib/airflow/xcom/__init__.py
@@ -15,3 +15,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from .backend import VineyardXCom


sighingnow (approved):
Thanks. That definitely a mistake.

sighingnow:
Hi @andtheysay,

I would like to let you know that we recently have made some update on the airflow integration, especially deploying on Kuberentes platform.

May I know if you still using vineyard to accelerate tasks in airflow? If vineyard is no longer in your consideration anymore, would you mind to take several minutes to discuss why vineyard didn't fit in your case?

Thanks in advance!

Issue #291, opened 2021-06-03
Vineyard-ctl: a command-line tool to inspect and operate the objects in vineyard cluster


sighingnow:
**Describe your problem**

There's a vineyard-cli tool in vineyard that provides limited functionalities on vineyard objects. A more powerful command-line tool vineyard-ctl is expected and at least the following features should be included:

+ supporting query, list and deletion on vineyard objects.
+ object migration request could be triggered.
+ monitoring the vineyard cluster and some useful metrics could be retrieved and displayed.
+ bonus: supporting scale in/out of vineyardd instances.

**Additional context**

N/A

chaitravi-ce:
I am interested in this project. I wanted to ask whether the project only includes integration of functions to command line or it also includes creating functions.

sighingnow:
The task includes two parts:

+ Implements a framework of cli application, and add operations on Objects (get/list/delete, etc.) to the cli.
+ Implements operation across many nodes: inspect the quota, drain a node, traigger migration, etc.

Spnetic-5:
Hey, @sighingnow, I would like to work on the proposal for this project & contribute to this project in OSPP Summer 2021. Currently, I'm unable to figure out how should I proceed with this project. Can you please share the approach to build the command-line tool vineyard-ctl?  Also, I have been facing some issues while installing the vineyard, can you please help me with it.
* Error says:
<p align="center"><img src="https://user-images.githubusercontent.com/66636289/121345949-1a875280-c943-11eb-897a-4a938167c12a.png" height = 250 width=600><p>

Hope to have the opportunity to discuss with you in more detail how I could contribute to this project.

sighingnow:
A module called "graph" in vineyard (for distributed graph data structures) requires https://github.com/alibaba/libgrape-lite, you could resolve the error by

1. disable that part by `cmake .. -DBUILD_VINEYARD_GRAPH=OFF`, or
2. install libgrape-lite first.

Kind regards,

chaitravi-ce:
Thank you @sighingnow, I faced the issue but it was resolved with the solution. But during make -j command my laptop hangs. Are there any hardware requirements for this process. I am using a laptop without GPU.

Spnetic-5:
Thanks @sighingnow it works, but now my pc has stuck while running `make -j`. Can you please provide solution to this, Can you please elaborate more on the approach to building the command-line tool vineyard-ctl.

sighingnow:
[...]

PR #467, opened 2021-08-28
Adds DeepCopy function to vineyard and Python SDK and adds it to vineyard-ctl

Rohan-cod:

What do these changes do?
-------------------------

This pull request is a combination of two commits.

`First commit`:

```
The first commit adds DeepCopy function to vineyard and the Python SDK.
The existing object migration code was used for the implementation.
```

`Second commit`:

```
The second commit updates vineyard-ctl to use the newly added
DeepCopy function when the user selects the --deep option under the copy command.
```

Part of #291.


sighingnow (approved):
LGTM.

PR #476, opened 2021-09-03
Adds memory_status option to query command in vineyard-ctl

Rohan-cod:

What do these changes do?
-------------------------

Added `memory_status` option to `query` **command** in `vineyard-ctl`. `memory_status` lets you print the **memory** used by a **vineyard object**.

Part of #291.

### DIFF
--- a/python/vineyard/cli.py
+++ b/python/vineyard/cli.py
@@ -106,6 +106,8 @@ def vineyard_argument_parser():
     query_opt.add_argument('--stdout', action='store_true', help='Get object to stdout')
     query_opt.add_argument('--output_file', type=str, help='Get object to file')
     query_opt.add_argument('--tree', action='store_true', help='Get object lineage in tree-like style')
+    query_opt.add_argument('--memory_status', action='store_true', help='Get the memory used by the vineyard object')
+    query_opt.add_argument('--detail', action='store_true', help='Get detailed memory used by the vineyard object')
 
     head_opt = cmd_parser.add_parser('head',
                                      formatter_class=argparse.RawDescriptionHelpFormatter,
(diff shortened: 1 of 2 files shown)


sighingnow:
Thanks for the quick pull request @Rohan-cod.

About the memory stat functionality, I think we could have an command line option, e.g., `--detail`, to control the behaviour. It says,

+ when `--detail=False`,  that is indeed what this pull request has implemented.
+ when `--detail=True`, I would like to see the memory usage of objects at different level/position in the hierarchical metadata. e.g., for dataframe, I would like to see output like

```
vineyard::DataFrame <"o23234sxsdfs">: 16Mi
   - vineyard::Tensor<int> <"o2342342sxx"> 8Mi
     - vineyard::Blob <"osdfs34...."> 8Mi
   - vineyard::Tensor<float> <"o342423dsxx"> 8Mi
     - ...
```

It would be useful for us to answer questions like "why the object takes so much of memory?" or "which part of the complex objects takes the majority of memory?".

Rohan-cod:
> have a command-line option, e.g., --detail, to control the behavior. 
> when `--detail=True`, I would like to see the memory usage of objects at different levels/positions in the hierarchical metadata. e.g., for the dataframe, I would like to see output like
> It would be useful for us to answer questions like "why the object takes so much of memory?" or "which part of the complex objects takes the majority of memory?".

Sure, I'll update the code accordingly.

python/vineyard/cli.py
```diff
@@ -452,20 +465,53 @@ def debug(client, args):
...
+            parent += f': {memory_used}'
+            memory_dict[parent] = memory_used
         tree.create_node(parent, parent)
     else:
-        new_parent = f'<{meta["typename"]}>:{meta["id"]}'
+        new_parent = f'{meta["typename"]} <{meta["id"]}>'
+        if memory:
+            memory_used = get_memory(meta["nbytes"])
+            new_parent += f': {memory_used}'
+            memory_dict[new_parent] = memory_used
```
siyuan0322, line 482: These lines can also be reorganized to avoid duplicates, like
```
new_parent = xxx
if memory:
   ...
tree.create_node(new_parent, new_parent, parent=parent)
parent = new_parent
```
Since the `create_node(self, tag=None, identifier=None, parent=None)` have this default `parent=None` you can always create a new parent and pass the old parent to it.

python/vineyard/cli.py
```diff
@@ -452,20 +465,53 @@ def debug(client, args):
...
+        if isinstance(meta[key], vineyard._C.ObjectMeta):
+            if str(meta[key]['typename']) == 'vineyard::Blob':
+                size = meta[key]['length']
+                total_bytes += size
+            else:
+                total_bytes += get_memory_used(meta[key])
+    return total_bytes
+
+
+def get_memory(nbytes):
```
siyuan0322, line 503: Rename to a meaningful name like `pretty_format_memory`

siyuan0322 (approved):
LGTM

PR #491, opened 2021-09-06
fix typename under gcc7.3 centos

siyuan0322:
Signed-off-by: siyuan0322 <siyuanzhang.zsy@alibaba-inc.com>


sighingnow:
Close as in favour of #492.

PR #498, opened 2021-09-13
Add the pair of init() and shutdown() methods to help new users to deploy vineyardd easily

### LINKED ISSUE #391: Add a vineyard.init() method to initialize a local vineyard instance after resolving configs.
Describe your problem
---------------------

Add a vineyard.init method which will launch the vineyardd server, according to configs that exist in the env, e.g., `VINEYARD_ETCD_ENDPOINT`, or `~/.vineyard.cfg`.

Additional context
------------------

andydiwenzhu:
What do these changes do?
-------------------------

In general, the init method handles three scenarios: no vineyardd exists; vineyardd already launched at an IPC socket;
mimic the local cluster mode. While the shutdown method can both close single or all the vineyardds.

Related issue number
--------------------

Fixes #391

### DIFF
--- a/python/vineyard/deploy/local.py
+++ b/python/vineyard/deploy/local.py
@@ -168,8 +168,8 @@ def init(proc_num=1, socket=None, **kw):
     
     it will launch a local vineyardd and return a connected client to the vineyardd.
 
-    If a vineyardd has been launched at **SOCKET**, pass the socket either by setting the
-    environment variable **VINEYARD_IPC_SOCKET** as the socket before head, or simply use:
+    If a specific **SOCKET** path is needed, pass the socket either by setting the
+    environment variable **VINEYARD_IPC_SOCKET** as the SOCKET before head, or simply use:
     
     .. code::
     
@@ -202,7 +202,10 @@ def init(proc_num=1, socket=None, **kw):
     etcd_endpoints = None
     etcd_prefix = f'vineyard_init_at_{time.time()}'
     for idx in range(proc_num):
-        ipc_socket = f'{socket}.{idx}' if socket else None
+        if proc_num > 1:
+            ipc_socket = f'{socket}.{idx}' if socket else None
+        else:
+            ipc_socket = socket
         try:
             client = connect(ipc_socket)
         except:
--- a/python/vineyard/deploy/local.py
+++ b/python/vineyard/deploy/local.py
@@ -206,14 +206,14 @@ def init(num_instances=1, **kw):
 
 def get_default_client():
     if not __default_instance_contexts:
-        raise ValueError("Vineyard has not been initialized, use vineyard.init()")
+        raise ValueError("Vineyard has not been initialized, use vineyard.init() to launch vineyard instances")
     clients = [__default_instance_contexts[k][1] for k in __default_instance_contexts]
     return clients if len(clients) > 1 else clients[0]
 
 
 def get_default_socket():
     if not __default_instance_contexts:
-        raise ValueError("Vineyard has not been initialized, use vineyard.init() to launch vineyard daemons")
+        raise ValueError("Vineyard has not been initialized, use vineyard.init() to launch vineyard instances")
     sockets = __default_instance_contexts.keys()
     return sockets if len(sockets) > 1 else sockets[0]
 
--- a/python/vineyard/deploy/local.py
+++ b/python/vineyard/deploy/local.py
@@ -16,6 +16,7 @@
 # limitations under the License.
 #
 
+import atexit
 import contextlib
 import logging
 import os
@@ -48,6 +49,8 @@ def start_vineyardd(etcd_endpoints=None,
         etcd_endpoint: str
             Launching vineyard using specified etcd endpoints. If not specified, vineyard
             will launch its own etcd instance.
+        etcd_prefix: str
+            Specify a common prefix to establish a local vineyard cluster.
         vineyardd_path: str
             Location of vineyard server program. If not specified, vineyard will use its
             own bundled vineyardd binary.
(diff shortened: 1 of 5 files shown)


python/vineyard/deploy/local.py
```diff
@@ -148,4 +153,67 @@ def start_vineyardd(etcd_endpoints=None,
...
+    '''
+    if socket is None:
+        socket = os.environ.get('VINEYARD_IPC_SOCKET', None)
+    try:
+        client = connect(socket)
+    except:
+        logger.info("Vineyard daemon on %s is not initialized", socket)
+        if etcd_prefix is not None:
+            kw['etcd_prefix'] = etcd_prefix
+            kw['etcd_endpoints'] = init.etcd_endpoints.get(etcd_prefix, None)
```
sighingnow, line 194: It looks very strange to resolve `etcd_endpoints` by `etcd_prefix`... Why the later is the key to the former?

python/vineyard/deploy/local.py
```diff
@@ -148,4 +153,67 @@ def start_vineyardd(etcd_endpoints=None,
...
+        vineyard.init(socket=SOCKET)
+
+    Sometimes we want to setup a local vineyard cluster consists of multiple vineyard daemons.
+    We can give the same **etcd_prefix** to each call of init:
+
+    .. code::
+    
+        client1 = vineyard.init(etcd_prefix='test')
+        client2 = vineyard.init(etcd_prefix='test')
+        client3 = vineyard.init(etcd_prefix='test')
```
sighingnow, line 182: Why they shares the etcd instance launched by previous vineyardd?

python/vineyard/deploy/local.py
```diff
@@ -148,4 +153,67 @@ def start_vineyardd(etcd_endpoints=None,
...
+        vineyard.init()
+    
+    it will launch a local vineyardd and return a connected client to the vineyardd.
+
+    If the user has already lauched a vineyardd with a **SOCKET**. She can either set VINEYARD_IPC_SOCKET
+    as the SOCKET, and then run vineyard.init(), or simply use:
+    
+    .. code::
+    
+        vineyard.init(socket=SOCKET)
```
sighingnow, line 173: Why these part? As we have discussed, DON'T make things so complicate....

sighingnow (changes requested):
BTW what will happen if users doesn't call shutdown? Image that user's code throw an exception and the python interpreter just exit, will there be orphan vineyardd process?

sighingnow:
Lets `init()` do things that "init()" should do, don't mix it with unrelated logic..

python/vineyard/deploy/local.py
```diff
@@ -134,7 +142,7 @@ def start_vineyardd(etcd_endpoints=None,
             raise RuntimeError('vineyardd exited unexpectedly with code %d, error is:\n%s' % (rc, err))
 
         logger.debug('vineyardd is ready.............')
-        yield proc, socket
+        yield proc, socket, etcd_endpoints
```
sighingnow, line 145: Please

1. check all use of `vineyard.deploy.local.start_vineyardd`  in the repo (including your previous notebook) to ensure that nothing is broken by this change
2. revisit the implementation of `vineyard.deploy.distributed.start_vineyardd` to ensure the function signatures are as consistent as possible.

python/vineyard/deploy/local.py
```diff
@@ -148,4 +156,86 @@ def start_vineyardd(etcd_endpoints=None,
             etcd_ctx.__exit__(None, None, None)  # pylint: disable=no-member
 
 
+def init(proc_num=1, socket=None, **kw):
```
sighingnow, line 159: `proc_num` -> "num_workers" or `num_instances`

as etcd is also a "proc".

python/vineyard/deploy/local.py
sighingnow: As we have discussed, `init` will always do launching, and `shutdown` will always do cleanup.

python/vineyard/deploy/local.py
```diff
@@ -148,4 +156,86 @@ def start_vineyardd(etcd_endpoints=None,
...
+            _, ipc_socket, etcd_endpoints = ctx.__enter__()
+            client = connect(ipc_socket)
+            init.contexts[ipc_socket] = (ctx, client)
+
+    return get_default_client()
+
+
+def get_default_client():
+    if not init.contexts:
+        raise ValueError("Vineyard has not been initialized, use vineyard.init()")
```
sighingnow, line 227: To "use `vineyard.init()` to luanch vineyard instances."

(to make it a complete sentence)

python/vineyard/deploy/local.py
```diff
@@ -148,4 +156,86 @@ def start_vineyardd(etcd_endpoints=None,
...
+    clients = [init.contexts[k][1] for k in init.contexts]
+    return clients if len(clients) > 1 else clients[0]
+
+
+@atexit.register
+def shutdown():
+    for ctx in reversed(init.contexts.keys()):
+        init.contexts[ctx][0].__exit__(None, None, None)
+    init.contexts = {}
```
sighingnow, line 236: Please do something like

```python
def shutdown()....

@atexit.register
def __shutdown_handler():
     try:
          shutdown()
     except Exception:
          pass
```

to avoid exceptions in `shutdown()` when `atexit` breaking other parts of user's program.

python/vineyard/deploy/local.py
```diff
@@ -148,4 +156,86 @@ def start_vineyardd(etcd_endpoints=None,
...
+
+
+init.contexts = {}
```
sighingnow, line 239: `init` to `_init` or `__init` to make it private.

sighingnow:
Hi @andydiwenzhu could you please revisit the pull request a bit further to address the comments above?

python/vineyard/deploy/local.py
sighingnow: `init` is a function.

I suggest to have a private global variable, rather than add custom attribute to function.

sighingnow:
Seems that the test failed on CI: https://github.com/v6d-io/v6d/pull/498/checks?check_run_id=3596553110#step:19:1971

sighingnow (approved):

PR #513, opened 2021-09-23
Trim BOM if found in beginning of file in LocalIOAdaptor

### LINKED ISSUE #512: Handle input files that are 'utf-8-sig' encoding
Text files from windows may contains a BOM bytes at the beginning, the arrow IO adaptor should drop these three bytes first before perform actual work.

siyuan0322:
Signed-off-by: siyuan0322 <siyuanzhang.zsy@alibaba-inc.com>

Fixes #512


siyuan0322:
Tested locally.

sighingnow:
Will it work when read table without headerrow?

sighingnow (approved):

siyuan0322:
> Will it work when read table without headerrow?

Yeah, handled that case in the else branch.

Issue #289, opened 2021-06-03
Improve the jemalloc-based allocator implementation.


sighingnow:
**Describe your problem**

The `TCACHE` is disabled in the jemalloc-base allocator, which, brings performance penality to the allocator and requires more work to be fixed.

Besides, currently the usage jemalloc seems has problems on Linux (Ubuntu 20.04) to reclaim allocated virtual memory 
address spaces.

**Additional context**

N/A.

sighingnow:
We have done for our goals.

Closing and thanks for the effort from @peilii !

PR #518, opened 2021-09-24
Arena test dev

peilii:
What do these changes do?
-------------------------

A thread-local arena demo test.

Related issue number
--------------------

Part of #289


sighingnow:
I would like to know:

1. each user thread will need to allocate an arena to itself (manually)?
2. it seems that from current code, thread won't reuse the arena, that is said, the number of arenas is decided by the number of active threads, rather than number of cores?
3. I'm interested in the overhead of looking up arena using the thread id

sighingnow:
It will be good if the current implementation could be reoriginaized as a `VineyarddAllocator` interfrace, and then how users could leverage the allocator could be more clear.

sighingnow:
I assume the `AllocateTask` is the equivalent interface for `malloc`, but seems that the implementation is not thread-safe.

peilii:
> I would like to know:
> 
> 1. each user thread will need to allocate an arena to itself (manually)?
> 2. it seems that from current code, thread won't reuse the arena, that is said, the number of arenas is decided by the number of active threads, rather than number of cores?
> 3. I'm interested in the overhead of looking up arena using the thread id

Thanks for the reply. This branch is just a simple demo of ArenaAllocator (which is implemented in another or). I would submit a new pull request.
1. Each thread would first request an arena before binding and allocating in that arena.
2.  When initializing ArenaAllocator, num cores arena would be created
3. I did run our benchmarks earlier today, and it seems like that the arena allocator would finish in 6379 ms, while the system default allocator would finish in 7080 ms. But it has some problems that can be improved. I assume the biggest problem now is how to reuse arenas after all arenas were bound.

sighingnow:
> I would submit a new pull request.

Thanks. Lets move to the new pull request.

> and it seems like that the arena allocator would finish in 6379 ms, while the system default allocator would finish in 7080 ms

Good to know the number.

PR #524, opened 2021-09-26
Arena allocator implement, to be completed

### LINKED ISSUE #289: Improve the jemalloc-based allocator implementation.

peilii:
What do these changes do?
-------------------------
An thread local arena allocator based on jemalloc.

Related issue number
--------------------

Fixes #289


sighingnow:
Hi @peilii, I have resolved those format & lint errors during compliation, I think this PR is a good startpoint and we can merge it then continue followup improvements.

As your concerns discussed on wechat,

> 将这个新 thread 和与它共享物理核的 thread 绑定在同一个 arena 上，这样 allocate 的时候还是不用加锁；

I think that is hard to implementation, as threads are not always bond to a fixed CPU core. Returns `nullptr` as error is ok at current stage.

> 但是小请求是用的 TCACHE_NONE，感觉这块还需要继续研究和完善一下，但是现在的方法可以正常跑起来

Its ok, IMO.

> 比之前好了很多，但是比 jemalloc 的表现还是差一点

Looks pretty promising 👍 

> 原本vineyard的malloc没有办法通过这个benchmark，在allocate一些内存后会返回nullptr，新的实现是可以通过的，但是在加上extent hooks之后速度比default jemalloc 和 system慢很多

I will try the newly added benchmark.

It is a known bug in the previous implementation, where the memory cannot be correctly reused. I guess the bug exists because we doesn't implements the "deallocate_hook", but I'm not sure about that.

> 新的实现是可以通过的，但是在加上extent hooks之后速度比default jemalloc 和 system慢很多

Do we have numbers?

sighingnow:
Merged.

peilii:
Hi @sighingnow ,

For the first problem, when all the arenas were assigned, I think we could try to bind each arena to store the thread ID of the last thread to allocate in the arena if required by future work.

As for the deallocate hook, I think the current implementation, which used the default deallocate hook works. But I am not quite sure why the original implementation ends without reusing allocated memories.

For numbers about performance, after adding extent hooks in the arena allocator, when running the alloc_test, it took 20093 ms on my Mac, while the original implementation took 24516 ms. When running jemalloc experiments benchmark, it took around 15 seconds to finish with 1 thread, around 200 seconds to finish with 4 threads, while the original implementation fails after several allocations in each case. It has some improvements compares to the original implementation, but the performance gap still exists, and I think the allocate hook has something to do with it.

Moreover, the usage and information about both benchmarks are described in the README.md file under their directories. Please let me know if there are any problems.

PR #528, opened 2021-09-27
Added Nvidia-Dali Builder and Resolver

chaitravi-ce:

What do these changes do?
-------------------------

* Added Builder and Resolver for Nvidia-Dali
* Added Tests

Related issue number
--------------------

Related to #280

### DIFF
--- a/python/vineyard/contrib/ml/tests/test_dali.py
+++ b/python/vineyard/contrib/ml/tests/test_dali.py
@@ -45,7 +45,7 @@ batch_size = 2
 num_threads = 4
 
 
-@pytest.mark.skip(dali is None, "nvidia-dali is not available")
+@pytest.mark.skip(dali is None, reason="nvidia-dali is not available")
 def test_dali_tensor(vineyard_client):
     @pipeline_def()
     def pipe():
--- a/python/vineyard/contrib/ml/tests/test_dali.py
+++ b/python/vineyard/contrib/ml/tests/test_dali.py
@@ -1,0 +1,60 @@
+#! /usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020-2021 Alibaba Group Holding Limited.
+#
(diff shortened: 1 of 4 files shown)


setup_ml.py
```diff
@@ -81,7 +81,7 @@ def find_ml_packages(root):
         "install": install_plat
     },
     zip_safe=False,
-    install_requires=['vineyard', 'tensorflow', 'xgboost', 'torch', 'mxnet'],
+    install_requires=['vineyard', 'tensorflow', 'xgboost', 'torch', 'mxnet', 'nvidia-dali-cuda110'],
```
sighingnow, line 84: I think we need to add a profile, namely `gpu`, where the `nvidia-dali-cuda110` is listed as a dependency, as it is not available on GPU.

python/vineyard/contrib/ml/tests/test_dali.py
```diff
@@ -0,0 +1,60 @@
...
+            register_dali_types(builder, resolver)
+            yield builder, resolver
+
+
+num_gpus = 1
+device_id = 0
+batch_size = 2
+num_threads = 4
+
+
```
sighingnow, line 44: Skip those tests if the dali package is not available.

setup_ml.py
chaitravi-ce: Yes dali requires cuda and nvidia drivers. I am not sure how we can add it here

sighingnow: Its all right, I can assistant to revise the PR.

chaitravi-ce: Thank You

sighingnow:
@chaitravi-ce I have added guard to check if dali is applicable in https://github.com/v6d-io/v6d/pull/528/commits/700da7fcea06d2ecc306635692b99028c33c4ad1, we can get this PR merged once CI turns green.

Thanks!

PR #560, opened 2021-10-28
Try implements vineyard data types as a subclass of existing C++ classes

sighingnow:
What do these changes do?
-------------------------

To show the potential for better/cleaner integration of vineyard with existing compute engines.

See `arrow_data_structure_v2_test.cc`, the `DoubleArray` that comes from vineyard (return by `client.get()`) can directly use any public method of `arrow::DoubleArray`.

Related issue number
--------------------

N/A


sighingnow:
Close as stale.

PR #563, opened 2021-10-31
Check graph is multigraph when `InitEdge` or `AddEdgeLabel`

acezen:
What do these changes do?
-------------------------

Related issue number
--------------------

N/A


modules/graph/fragment/arrow_fragment.h
```diff
@@ -1878,6 +1881,8 @@ class ArrowFragment
       }
     }
 
+    LOG(INFO) << "existed_parallel_edge=" << existed_parallel_edge;
```
sighingnow, line 1884: `VLOG(2)` would be better here.

sighingnow:
Please rebase and sign-off your commits as well.

modules/graph/fragment/arrow_fragment.h
acezen: removed

acezen:
cc/ @siyuan0322

PR #585, opened 2021-11-15
MacOS: wheel built with static linkage and `hidden` visibility.

sighingnow:


sighingnow:
To make sure that `.global_init` doesn't change the visibility of symbols.

PR #605, opened 2021-12-03
Enable auto dlopen when initialize the io factory.

sighingnow:
What do these changes do?
-------------------------

Load extra io adaptors from environment variable `VINEYARD_OTHER_IO_ADAPTORS`.

Related issue number
--------------------

n/a.

### DIFF
--- a/modules/io/io/io_factory.cc
+++ b/modules/io/io/io_factory.cc
@@ -15,6 +15,8 @@ limitations under the License.
 
 #include "io/io/io_factory.h"
 
+#include <dlfcn.h>
+
 #include <limits.h>
 #include <stdlib.h>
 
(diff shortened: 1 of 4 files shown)


modules/io/io/io_factory.cc
```diff
@@ -27,11 +29,28 @@ limitations under the License.
...
+
 namespace vineyard {
 
-void IOFactory::Init() {}
+void IOFactory::Init() {
+  // load other io adaptors using dlopen
+  auto other_io_adaptors = read_env("VINEYARD_OTHER_IO_ADAPTORS");
+  std::vector<std::string> adaptors;
+  ::boost::split(adaptors, other_io_adaptors,
+                 ::boost::is_any_of(std::string(1, ':')));
```
siyuan0322, line 44: Why not just use `is_any_of(":")`

siyuan0322 (approved):

PR #682, opened 2022-03-20
Add a context protocol type to  support dynamic::kNullType and np.void or arrow::null

acezen:
What do these changes do?
-------------------------

Related issue number
--------------------

Fixes #issue number

### DIFF
--- a/modules/graph/utils/context_protocols.h
+++ b/modules/graph/utils/context_protocols.h
@@ -75,7 +75,9 @@ struct TypeToInt<std::string> {
 };
 
 inline int ArrowDataTypeToInt(const std::shared_ptr<arrow::DataType>& type) {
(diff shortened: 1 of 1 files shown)


modules/graph/utils/context_protocols.h
```diff
@@ -29,6 +29,11 @@ struct TypeToInt {
   static constexpr int value = -1;
 };
 
+template <>
+struct TypeToInt<void> {
+  static constexpr int value = 0;
+};
```
sighingnow, line 35: Add a check in `ArrowDataTypeToInt` as well, it should be `arrow::null()`

sighingnow (changes requested):

sighingnow (approved):

PR #694, opened 2022-03-30
split table_shuffler_beta impl to cc file

### LINKED ISSUE #693: Split the function definitions in `table_shuffler_beta.h` to a `.cc` file
Describe your problem
---------------------

as titled

zhanglei1949:
Split impl away from header to avoid multiple definition when linking.

What do these changes do?
-------------------------

Related issue number
--------------------

Fixes #693

### DIFF
--- a/modules/graph/utils/table_shuffler_beta.cc
+++ b/modules/graph/utils/table_shuffler_beta.cc
@@ -1,0 +1,282 @@
+/** Copyright 2020-2021 Alibaba Group Holding Limited.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


modules/graph/utils/table_shuffler_beta.cc
```diff
@@ -0,0 +1,151 @@
...
+                          std::shared_ptr<arrow::Array> array,
+                          const std::vector<int64_t>& offset) {
+  auto* ptr = std::dynamic_pointer_cast<arrow::LargeListArray>(array).get();
+  for (auto x : offset) {
+    arrow::LargeListArray::offset_type length = ptr->value_length(x);
+    arc << length;
+    auto value = ptr->value_slice(x);
+    serialize_selected_typed_items<T>(arc, value);
+  }
+}
```
sighingnow, line 48: The implementation of template functions should still be placed in `.h`.

sighingnow (changes requested):

PR #704, opened 2022-04-08
Remove write msg mutex for SocketConnection.

septicmk:
Only one thread can access the `write_msgs_`, thus we do not need the lock.


sighingnow (changes requested):
I suggest remove the `write_msgs_` queue as well.

PR #710, opened 2022-04-20
Correctly find vineyard package on Debian with multiarch support

lidongze0629:
What do these changes do?
-------------------------

Fixes CI failed on [GraphScope](https://github.com/alibaba/GraphScope/runs/6083691703?check_suite_focus=true)

Follow [CMAKE Doc](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html)

> LIBDIR: object code libraries (lib or lib64) 
                 On Debian, this may be lib/multiarch-tuple when CMAKE_INSTALL_PREFIX is /usr

I'm not sure whether has a better way to fix it. Welcome any comments.

Related issue number
--------------------

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,6 +168,9 @@ else()
     set(HAVE_FLAG_STD_CXX14 TRUE)
 endif()
 
+set(CMAKE_INSTALL_DIR lib/cmake/vineyard CACHE STRING
+    "The subdirectory where CMake package config files should be installed")
+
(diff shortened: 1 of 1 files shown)


lidongze0629:
Refer [folly](https://github.com/facebook/folly/blob/main/CMakeLists.txt#L427) and [cmake](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#example-generating-package-files)

PR #711, opened 2022-04-21
Fix incorrect value of partition size in `GlobalDataFrame`

siyuan0322:
Signed-off-by: siyuan0322 <siyuan0322@gmail.com>


sighingnow (approved):
Nice catch!

PR #743, opened 2022-05-18
add a dep when compile on macos

waruto210:
Signed-off-by: waruto210 <wmc314@outlook.com>

What do these changes do?
add a dependency when compile on macOS

add a dependency when compile on macOS

no issue


sighingnow (approved):

sighingnow:
Merged.

Thank you @waruto210

Issue #728, opened 2022-05-10
Accept both recordbatch stream and dataframe stream in the fragment loader


sighingnow:
Describe your problem
---------------------

Refactor the stream reader to formulate some untilities to load chunks from streams, both dataframe streams and recordbatch streams, as they serve a very similar propuse.

See also current implementation: https://github.com/v6d-io/v6d/blob/main/modules/graph/loader/arrow_fragment_loader.h#L57

Additional context
------------------

PR #776, opened 2022-06-06
Fixes the compatibility issue of stream implementation

sighingnow:
What do these changes do?
-------------------------

Fixes the resolver/builder for dataframe stream and recordbatch stream and make sure their semantic matches between C++ and Python.

Related issue number
--------------------

N/A


sighingnow:
See also: https://github.com/v6d-io/v6d/issues/728

Issue #739, opened 2022-05-14
UsageTracker: track the least-recently-used blobs in vineyard server.


septicmk:
As titled.
- [x] #765 
- [ ] (TODO) Apply LRU on cold blobs.

ZjuYTW:
I am about to add this feature. And the following is a draft-plan, appreciate for any advice!

First achieve a singleton LRU class
```
class LRU {
    using lru_iter = std::list<ObjectID>::iterator;
    using clru_iter = std::list<ObjectID>::const_iterator;

    public:
    static LRU* GetInstance();

    void Refer(ObjectID obj_id) const;

    clru_iter begin() const; // used for iterating through the list

    private:
    LRU() {} // NOLINT

    std::list<ObjectID> cache_;
    std::unordered_map<ObjectID, lru_iter> iter_table_;
    size_t max_size_; // Maybe we need a threshold to keep track of those hot blob

    static LRU* lru_;
  }
```

Then we should call Lru::refer every time we try to visit blob
```
const std::shared_ptr<arrow::Buffer>& Blob::Buffer() const {
  ...
  Lru::GetInstance()->refer(this->id_);
  return buffer_;
}
```

A obvious drawback is that we can't refresh the lru list everytime the buffer_ is used...

sighingnow:
Ke @septicmk  may you give some comment on the proposal above?

Thanks!

septicmk:
Hi @ZjuYTW, An LRU cache may not be enough for later spilling. For example, we should not spill an object already in use. I suggest we should track the usage of every object, and only maintain LRU caches for unused objects. This means we should record something like “reference count” and only put the objects that "reference count==0"  into your LRU cache.

FYI, we already have a usage tracker on the client-side to maintain the reference count (https://github.com/v6d-io/v6d/blob/main/src/client/client.h#L122), we could have something similar on the server-side.

ZjuYTW:
@septicmk  Sry for the lag on response (little bit busy in previous weeks). As you mentioned, we need to keep track of all objects in the server side.
And there are two BulkStore type in vineyard_server, and seems we don't need to implement UsageTrack on `BulkStore` ? 
For `PlasmaBulkStore`, because already have a `reference counter` CRTP to tracker reference of object, we may add the LRU cache in the PlasmaBulkStore's as a private member to track the unused objects.

[...]

PR #794, opened 2022-06-19
Support lru in `ColdObjectTracker`

### LINKED ISSUE #724: Implements spill and checkpoint functionalities in Vineyard
Describe your problem
---------------------

Vineyard is in-memory data manager for bigdata computation workflows. Vineyard shares distributed datasets (e.g., tensors, dataframes, graphs) across many machines to enable zero-copy data sharing between distributed compute engines. In some read-world cases, the data may sometimes exceed the available memory size, and can be swapped to disks (or remote storage like oss or s3) to temporarily release and memory for other jobs and swapped it back to memory again when been required. It is the so-called "spill" process.

In this task the candidate is responsible for implementing such a spill functionality in Vineyard, with a relative reasonable and smart policy (co-designed with the mentor) that can select proper objects to spill, e.g., it is unsane to spill an object that is current in use. Based the support for spill and reload, we can archive the checkpoint functionality which could temporarily dump the whole data in vineyard and reload it back some time later. The checkpoint functionality is the basis for fault-tolerance.

SubTasks
---------
+ [x] A warmup task to get familiar with blob store: 
   - [x] #739  
+ [x] Implements the support for spill in vineyard, local disks, and remote object storage must be supported via a unified and extensible interface.
   - [x] #740 
+ [ ] Implements the support for checkpoint in vineyard by reusing some functionalities in spill.
   - [ ] #243

Additional context
------------------

This issue is part of our [OSPP 2022](https://summer-ospp.ac.cn/#/org/orgdetail/bb1d0787-db22-4203-b15d-a1f1bf12548b/)

ZjuYTW:
What do these changes do?
-------------------------

In this PR, I will build a LRU for ColdObjectTracker, to keep track of the oldest unused `Obj ID`.
Mainly prepare for later implementation of `spill`.

In detail, I sketched a LRU by a `doubly-linked list` and  `hash table` -- A classic implementation of LRU. Due to the limitation of currently used cpp version (C++14), the synchronize mechanism I firstly choosed was `std::shared_mutex` in C++17, then I changed to `std:: shared_timed_mutex`, but they should have no difference in our scenario.  [see here](https://stackoverflow.com/questions/40207171/why-shared-timed-mutex-is-defined-in-c14-but-shared-mutex-in-c17)

Related issue number
--------------------

Fixes #724(partially) and https://github.com/v6d-io/v6d/issues/739

TODOs
-------

- [ ]  Add UnitTest 

- [x] Encounter a problem when I attempt to run test on my server. Plz kindly refer to disscussion ([Disccussion here](https://github.com/v6d-io/v6d/discussions/741))

Updates
--------
6.21 Add a basic unit test. Have a semantic problem with LRU -- Do we need update the position(like move it to the head position) for the blob when `CheckExists(id)` is called? @septicmk

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -774,6 +774,10 @@ if(BUILD_VINEYARD_TESTS)
             target_compile_options(${T_NAME} PRIVATE "-fno-access-control")
         endif()
 
+        if(${T_NAME} STREQUAL "lru_test")
+            target_link_libraries(${T_NAME} PRIVATE TBB::tbb)
+        endif()
+
         if(${T_NAME} STREQUAL "allocator_test"
                 OR ${T_NAME} STREQUAL "arena_test"
                 OR ${T_NAME} STREQUAL "jemalloc_test")
--- a/src/server/memory/usage.h
+++ b/src/server/memory/usage.h
@@ -174,11 +174,16 @@ class DependencyTracker
 template <typename ID, typename P, typename Der>
 class ColdObjectTracker
     : public DependencyTracker<ID, P, ColdObjectTracker<ID, P, Der>> {
+ public:
   /*
    * @brief LRU is a tracker of least recent used blob ID, it has two methods:
-   * - `Ref(ID id)` Add the id if not exists, or refresh usage if exists
+   * - `Ref(ID id)` Add the id if not exists. (Actually here we shouldn't expect
+   *    a redundant Ref, because no Object will be insert twice). But in current
+   *    implementation, we will overwrite the previous one.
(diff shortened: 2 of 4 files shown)


sighingnow (changes requested):
Please rebase and sign-off your commit.

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

sighingnow:
Vineyard currently built with `C++14`, and we currenly have no plan to upgrade to C++17 for backwards compatibility reason.

sighingnow (approved):

PR #839, opened 2022-07-11
make all GenerateBlobID use rand generator

### LINKED ISSUE #837: Change `GenerateBlobID` to make it compatible for address reuse
Describe your problem
---------------------

In the #740  we gonna spill the object onto disk and free the memory. So in a chance that we will get a complete same address, and according to our current logic, it will be mapped into **the same ObjectID**, which obviously wrong...

Maybe we could redesign a pattern for this... (like use random number generator or standard timestamp)

Additional context
------------------
Potential Problem:
Not sure if we change this logic, then there will be a problem for online service, like, can't find the right BlobID?

ZjuYTW:
Signed-off-by: ZjuYTW <wangyitao999@outlook.com>

What do these changes do?
-------------------------
Use random generator for `Generate BlobID`

Related issue number
--------------------

Fixes #837 #777

### DIFF
--- a/src/server/memory/memory.cc
+++ b/src/server/memory/memory.cc
@@ -236,8 +236,7 @@ Status BulkStoreBase<ID, P>::Get(std::vector<ID> const& ids,
 template <typename ID, typename P>
 Status BulkStoreBase<ID, P>::Delete(ID const& object_id) {
   if (object_id == EmptyBlobID<ID>() ||
-      object_id == GenerateBlobID<ID>(reinterpret_cast<void*>(
-                       std::numeric_limits<uintptr_t>::max()))) {
+      object_id == GenerateBlobID<ID>(std::numeric_limits<uintptr_t>::max())) {
     return Status::OK();
   }
   typename object_map_t::const_accessor accessor;
@@ -342,8 +341,7 @@ Status BulkStoreBase<ID, P>::PreAllocate(const size_t size) {
   }
 
   // insert a special marker for obtaining the whole shared memory range
-  ID object_id = GenerateBlobID<ID>(
-      reinterpret_cast<void*>(std::numeric_limits<uintptr_t>::max()));
+  ID object_id = GenerateBlobID<ID>(std::numeric_limits<uintptr_t>::max());
   int fd = -1;
   int64_t map_size = 0;
   ptrdiff_t offset = 0;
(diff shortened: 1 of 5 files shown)


src/server/memory/memory.cc
```diff
@@ -420,8 +418,7 @@ Status BulkStoreBase<ID, P>::RemoveOwnership(
     std::set<ID> const& ids, std::map<ID, P>& successed_id_to_size) {
   for (auto id : ids) {
     if (id == EmptyBlobID<ID>() ||
-        id == GenerateBlobID<ID>(reinterpret_cast<void*>(
-                  std::numeric_limits<uintptr_t>::max()))) {
+        id == GenerateBlobID<ID>(std::numeric_limits<uintptr_t>::max())) {
```
sighingnow, line 421: Use `InvalidObjectID()`.

src/server/memory/memory.cc
```diff
@@ -342,8 +341,7 @@ Status BulkStoreBase<ID, P>::PreAllocate(const size_t size) {
   }
 
   // insert a special marker for obtaining the whole shared memory range
-  ID object_id = GenerateBlobID<ID>(
-      reinterpret_cast<void*>(std::numeric_limits<uintptr_t>::max()));
+  ID object_id = GenerateBlobID<ID>(std::numeric_limits<uintptr_t>::max());
```
sighingnow, line 344: Use `InvalidObjectID()`.

src/server/memory/memory.cc
```diff
@@ -236,8 +236,7 @@ Status BulkStoreBase<ID, P>::Get(std::vector<ID> const& ids,
 template <typename ID, typename P>
 Status BulkStoreBase<ID, P>::Delete(ID const& object_id) {
   if (object_id == EmptyBlobID<ID>() ||
-      object_id == GenerateBlobID<ID>(reinterpret_cast<void*>(
-                       std::numeric_limits<uintptr_t>::max()))) {
+      object_id == GenerateBlobID<ID>(std::numeric_limits<uintptr_t>::max())) {
```
sighingnow, line 239: Use `InvalidObjectID()`.

sighingnow (approved):

src/server/memory/memory.cc
ZjuYTW: you mean `GenerateBlobID<ID>(InvalidObjectID())` ?

ZjuYTW: sry for my mistake, the correct semantic for `GenerateBlobID<ID>(std::numeric_limits<uintptr_t>::max())` is to generate `0xFFFFFFFFFFFFFFFF` which is `0x8000000000000000UL | 0x0x7FFFFFFFFFFFFFFFUL`.

Now fixed in my 3rd commit, plz check

septicmk (approved):

Issue #911, opened 2022-08-11
Remove the unused `VineyardJob` source code and controller in k8s/ directory


sighingnow:
Describe your problem
---------------------

It is not used anymore and has been proved as a failure experiment. I will clean up this part for preparation about our next k8s development stage.

Additional context
------------------

Will done before the end of this week.

PR #917, opened 2022-08-12
Infer the job replicas from the pods's ownership references

sighingnow:
What do these changes do?
-------------------------

as titled. Try our best to infer the replicas from the pod's owner references.

Related issue number
--------------------

Part of #911


dashanji (approved):
LGTM

PR #931, opened 2022-08-23
Add e2e test for scheduler-plugin.

### LINKED ISSUE #916: Improve the scheduler plugin for the actual production environment
Describe your problem
---------------------

The current scheduler plugin isn't enough to meet the needs of the actual production environment, we should improve the scheduler plugin to make it easy for end-users to use the vineyard's ability in the k8s environment.

### What is the problem:

1. In the actual production environment, users need to align an order for a workflow.
2. At present, the CRD `GlobalObject` and `LocalObject` are considered observed resources, and we should get the workload that produces it through `annotation` or `label` and add ownereference for them.
3. There isn't a demo for users.

### Todo List:
- [ ] ~~Add some annotations for users to setup the workflow order etc.~~
- [ ] ~~Add  ownereference for `GlobalObject` and `LocalObject`.~~
- [x] Containerize the current [demo](https://v6d.io/notes/ml.html) so that it can be deployed in the k8s cluster and used in the operator.
- [x] Add an e2e test for the operator.
- [ ] ~~Add user manual.~~

dashanji:
What do these changes do?
-------------------------

* Refactor the scheduler-plugin.
* Add a simple demo for e2e test.
* Introduce skywalking-infra-e2e to test vineyard operator in CI.

Related issue number
--------------------

Fixes #916


sighingnow (approved):

.pylintrc
```diff
@@ -60,10 +60,7 @@ confidence=
 # no Warning level messages displayed, use "--disable=all --enable=classes
 # --disable=W".
 disable=abstract-method,
-        apply-builtin,
         attribute-defined-outside-init,
-        backtick,
-        bad-continuation,
```
sighingnow, line 66: Why removing this lines?

PR #939, opened 2022-08-24
Upgrage crdVersion to v1 and add multigroup

dashanji:
What do these changes do?
-------------------------

As titled. We can add new CRD smoothly then.


sighingnow (approved):
LGTM.

Thanks for doing this!

PR #940, opened 2022-08-24
add cache manager to fuse

liusitan:
[liusitan](https://github.com/liusitan)
liusitan commented [1 minute ago](https://github.com/bobhan1/v6d/pull/1#issue-1350124739)
This addition can manage the use of the fuse process, with default set to 1 GB.

However, the issue with this branch is that the pytest on the fuse manager seems not working, but it is testable when running fuse/tests/fuse_wild.py.


sighingnow:
@liusitan Ping me when this pull request is _ready for review_.

sighingnow:
Hi @liusitan, do you still have bandwidth to finish this pull request?

Thanks!

liusitan:
yes, but I have several concerns about the implementation detail that I end up not figuring out， especially when I run the performant test, I will leave a more detailed message tomorrow. It involves the plasma mechanisms.

PR #951, opened 2022-08-26
Add Vineyardd CRD and Controller

dashanji:
What do these changes do?
-------------------------

1. Add Vineyardd CRD and Controller.
2. Add a label `job:none` when creating the globalobject and localobject in the vineyardd.
3. Add serviceaccount、clusterrole and clusterrolebinding for vineyardd.
4. Enable Webhook configuration.
5. Add e2e test for Vineyardd CRD.


sighingnow (approved):

sighingnow:
@dashanji Ping me when this pull request is ready for merge.

dashanji:
Copy that, I will add the checkpoint mechanism in the pr so that we can release a usable vineyard operator in the next PR.

dashanji:
@sighingnow Ready.

sighingnow (changes requested):
Why we need the top-level `go.sum`?

dashanji:
> Why we need the top-level `go.sum`?

Sorry, I add it by accident. =.=

k8s/apis/k8s/v1alpha1/vineyardd_types.go
```diff
@@ -0,0 +1,193 @@
...
+
+// MetricConfig holds all metrics configuration
+type MetricConfig struct {
+	// represent the metric's image
+	// +kubebuilder:validation:Optional
+	Image string `json:"image,omitempty"`
+	// the policy about pulling image
+	// +kubebuilder:validation:Optional
+	ImagePullPolicy string `json:"imagePullPolicy,omitempty"`
+}
```
sighingnow, line 33: It is wired to call `Image` and `ImagePullPocliy` as "metric".

k8s/apis/k8s/v1alpha1/vineyardd_types.go
```diff
@@ -0,0 +1,193 @@
...
+	Name string `json:"name,omitempty"`
+	// the path of serializetion
+	// +kubebuilder:validation:Optional
+	Path string `json:"path,omitempty"`
+	// the configuration of the serialization PV
+	// +kubebuilder:validation:Optional
+	PersistentVolumeSpec corev1.PersistentVolumeSpec `json:"persistentVolumeSpec,omitempty"`
+	// the PersistentVolumeClaimSpec of the serializtion
+	// +kubebuilder:validation:Optional
+	PersistentVolumeClaimSpec corev1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
```
sighingnow, line 107: If I understand it correctly, `serialization` is a user-triggerred operator (e.g., invoke `vineyard.io.serialize()` in user's workload).

Why do we need add such a configuration to the deployment of vineyardd itself?

The serialization volume should be mounted along with the workload, rather than vineyardd.

k8s/templates/vineyardd/serialize-pv.yaml
```diff
@@ -0,0 +1,70 @@
...
+    {{- end }}
+  {{- end }}
+  {{- if $spec.StorageClassName }}
+  storageClassName: {{ $spec.StorageClassName }}
+  {{- end }}
+  {{- if $spec.VolumeMode }}
+  volumeMode: {{ $spec.VolumeMode }}
+  {{- end }}
+  {{- end }}
+{{ end }}
```
sighingnow, line 70: Serialzation shouldn't be part of vineyardd deployment.

src/server/util/kubectl.cc
```diff
@@ -19,6 +19,7 @@ limitations under the License.
 #include <memory>
 #include <string>
 #include <vector>
+#
```
sighingnow, line 22: Remove this empty line.

src/server/util/kubectl.cc
```diff
@@ -101,6 +117,14 @@ static std::string generate_global_object(
...
                     "\nmetadata:"
                     "\n  name: " + object_id +
+                    "\n  namespace: " + namespace_ +
+                    "\n  labels:"
+                    "\n    job: none"
+                    "\n  ownerReferences:"
+                    "\n    - apiVersion: v1"
+                    "\n      kind: Pod"
+                    "\n      name: " + pod_name +
+                    "\n      uid: " + uid +
```
sighingnow, line 127: I was imaging the owner reference of those local objects / global objects should be the `Vineyardd` deployment, rather than some single pod.

Because if one node failure, other node will do some garbage collection to deleted the local objects that been placed on that node, as well as global objects that requires those local objects. During this process those CRDs will be deleted by vineyardd.

We want to the owner reference to handle cases that when the vineyardd cluster (vineyardd CRDs) been deleted as a whole, there would be no role to do the garbage collection work, thus we need to leverage the owernship reference to remove those local / global object CRDs.

sighingnow (changes requested):
@dashanji I have leaf some comments.

k8s/apis/k8s/v1alpha1/vineyardd_types.go
dashanji: The `Image` and `ImagePullPocliy ` is the field in the Metric Container, and it is used to differentiate the vineyardd's configuration.

src/server/util/kubectl.cc
dashanji: Got it.

k8s/apis/k8s/v1alpha1/vineyardd_types.go
dashanji: Make sense to me, what I imagined earlier is that both vineyardd and user pods need to recognize this `serialization 
 path` at the same time. I will delete the field later.

src/server/util/kubectl.cc
dashanji: Yep, in such a scene, using the pod as ownerreference is unsuitable. I will add a `kubectl get` API  to query the `vineyardd` deployment in the `kubectl.cc`.

sighingnow: No need to such an API.

If those vineyardd pod is managed by a `Vineyardd` CRD, you could add the UID of `Vineyardd` CRD to the environment variable of those pods, I think.

dashanji:
@sighingnow Thanks for reviewing, all comments are done.

sighingnow (approved):

dashanji:
Thanks for the guidance, teacher tao. = =

PR #1006, opened 2022-09-28
Fix bug in vineyard-operator ci

dashanji:
As titled.

### DIFF
--- a/.github/workflows/vineyard-operator.yaml
+++ b/.github/workflows/vineyard-operator.yaml
@@ -93,7 +93,7 @@ jobs:
         if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v') && github.repository == 'v6d-io/v6d' }}
         run: |
           cd k8s
-          sudo make docker-push IMG=${IMG}:${{ github.sha }}
+          sudo make docker-build IMG=${IMG}:${{ github.sha }}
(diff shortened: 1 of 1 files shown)


sighingnow (approved):

.github/workflows/vineyard-operator.yaml
```diff
@@ -93,7 +93,7 @@ jobs:
         if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v') && github.repository == 'v6d-io/v6d' }}
         run: |
           cd k8s
-          sudo make docker-push IMG=${IMG}:${{ github.sha }}
+          sudo make docker-build IMG=${IMG}:${{ github.sha }}
           sudo docker tag ${IMG}:${{ github.sha }} vineyardcloudnative/vineyard-operator:latest
           sudo docker tag ${IMG}:${{ github.sha }} vineyardcloudnative/vineyard-operator:${{ steps.tag.outputs.TAG }}
           sudo docker push vineyardcloudnative/vineyard-operator:latest
```
sighingnow, line 99: Don't push to dockerhub from CI, as we didn't login into dockerhub here (and we won't do that in current plan). Pushing to github packages is enough.

sighingnow (changes requested):

.github/workflows/vineyard-operator.yaml
dashanji: Copy.

PR #1017, opened 2022-10-14
Add Dockerfile for vineyard python-dev

dashanji:
What do these changes do?
-------------------------

The pr is to add a dockerfile for building a minimal python client image for vineyard.

Why we need this?
-------------------------

During development, if we update the code related to the vineyard python client, I think it's better to have an existing dockerfile to build the python-dev image.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -10,7 +10,7 @@ PYTHON_DEV_REGISTRY	:= docker.pkg.github.com/v6d-io
 PYTHON_DEV_IMAGE	:= v6d/vineyard-python-dev
 PYTHON_DEV_TAG		:= latest
 
-all: builder docker-build build-python-dev docker-push
+all: builder docker-build docker-push
 
 builder: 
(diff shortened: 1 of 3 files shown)


docker/Makefile
```diff
@@ -6,7 +6,11 @@ ALPINE_REGISTRY		:= docker.pkg.github.com/v6d-io
 ALPINE_IMAGE		:= v6d/vineyardd
 ALPINE_TAG			:= alpine-latest
 
-all: builder docker-build docker-push
+PYTHON_DEV_REGISTRY	:= docker.pkg.github.com/v6d-io
+PYTHON_DEV_IMAGE	:= v6d/vineyard-python-dev
+PYTHON_DEV_TAG		:= latest
+
+all: builder docker-build build-python-dev docker-push
```
sighingnow, line 13: The `build-python-dev` shouldn't be included in `all`.

sighingnow (changes requested):

docker/Makefile
dashanji: Done

sighingnow (approved):

PR #1021, opened 2022-10-19
Add Docker Hub label in the root readme

dashanji:
As titled. The docker hub link is important for users, I think.


sighingnow (approved):

dashanji:
Sorry, I forgot about it.....

PR #1046, opened 2022-11-02
Delete the Name field in globalobject and localobject's API

dashanji:
Delete the Name field as we never use them. E.g.

```shell
kubectl get globalobjects -n vineyard-system
NAME                ID                  NAME   SIGNATURE           TYPENAME
o001c87014cf03c70   o001c87014cf03c70          s001c87014cf03262   vineyard::Sequence
```

```shell
 kubectl get localobjects -n vineyard-system 
NAME                ID                  NAME   SIGNATURE           TYPENAME                  INSTANCE   HOSTNAME
o001c87014ca81924   o001c87014ca81924          s001c87014ca80acc   vineyard::Tensor<int64>   1          kind-worker2
o001c8729e4590626   o001c8729e4590626          s001c8729e458f47a   vineyard::Tensor<int64>   2          kind-worker3
```

### DIFF
--- a/charts/vineyard-operator/templates/crds.yaml
+++ b/charts/vineyard-operator/templates/crds.yaml
@@ -47,9 +47,6 @@ spec:
     - jsonPath: .spec.id
       name: Id
       type: string
-    - jsonPath: .spec.name
-      name: Name
-      type: string
     - jsonPath: .spec.signature
       name: Signature
       type: string
@@ -75,8 +72,6 @@ spec:
                   type: string
                 type: array
               metadata:
-                type: string
-              name:
                 type: string
               signature:
                 type: string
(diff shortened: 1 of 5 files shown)


sighingnow:
Objects do have "names".

dashanji:
I think the `.metadata.name` can indicate it.

sighingnow:
> I think the `.metadata.name` can indicate it.

`.metadata.name` is expected to be unique and globally consistent so it would be (and currently is) object ID, won't be "name", as `name` property is optional and assigned by `client.put_name()`.

Not all objects have "name".

dashanji:
Make sense to me, thanks for the detail. Closed.

PR #1057, opened 2022-11-10
Add default value to vineyard deployment, remove ad-hoc filling and fixes typos.

sighingnow:
What do these changes do?
-------------------------

as titled. Moving default values to api types and remove the ad-hoc filling in both webhooks and template YAMLs. Rewrite some examples of `kind: Vineyardd` to make is as concise as possible.

Related issue number
--------------------

N/A


dashanji (approved):

k8s/templates/vineyardd/deployment.yaml
```diff
@@ -167,19 +142,15 @@ spec:
...
       - name: run
         hostPath:
-          {{- if .Spec.Config.Socket }}
-          path: {{ .Spec.Config.Socket }}
-          {{ else }}
-          path: /var/run/vineyard-{{ .Namespace }}-{{ .Name }}
-          {{- end }}
-      {{- if .Spec.Config.SpillConfig.Name }}
-      - name: {{ .Spec.Config.SpillConfig.Name }}
+          path: {{ .Spec.Socket }}
```
dashanji, line 149: I think the e2e test failed at this, it should be `/var/run`.

dashanji: And all path `/var/run/vineyard-vineyard-system-vineyardd-sample` should be changed to the default socket path `/var/run`. https://github.com/v6d-io/v6d/blob/71c0afedef8aceffbf3bd1ed3db390741ba689a0/k8s/test/e2e/workflow-job1.yaml#L50

FYI, @sighingnow

PR #1065, opened 2022-11-16
Use the helmify tool to autogenerate operator helm charts and update relevant e2e test and doc.

dashanji:
What do these changes do?
-------------------------
* Use the [helmify](https://github.com/arttor/helmify) to generate the vineyard operator helm charts automatically.
* Add the autogenerated helm chart e2e test.
* Update the relevant doc.


charts/vineyard-operator/templates/_helpers.tpl
```diff
@@ -50,3 +50,13 @@ app.kubernetes.io/name: {{ include "vineyard-operator.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
```
sighingnow, line 52: Why changing the file permission? Looks strange to have `x` bit for a tpl file.

dashanji: These file's permission is autogenerated by the `helmify` first and I update the permission manually. Besides, the helmify doesn't overwrite the previous files and it will not happen again.

PR #1095, opened 2022-12-08
Enhance implementation of fragment to accelerate the compilation

sighingnow:
What do these changes do?
-------------------------

Explicitly instantiatize the fragment, vertexmap, and loaders for `int32_t`, `int64_t` and `std::string`, which should be enough for `ArrowFragment`.

Related issue number
--------------------

N/A


siyuan0322 (approved):
Splendid!

PR #1113, opened 2022-12-23
Fixes parsing flags when using higher version of apache-arrow

sighingnow:
What do these changes do?
-------------------------

Fixes a include error with arrow 6.0 as well.


siyuan0322 (approved):
👍

PR #1138, opened 2023-01-10
Implements accmulate support for vineyard.io.open()

### LINKED ISSUE #1136: Vineyard I/O: support returns an object instead of a stream
Describe your problem
---------------------

Current `vineyard.io.open()` can only returns a stream, not a dataframe, that causing some differences with the behavior of things like `pd.read_csv()`.

An option `accumulate` would be needed to make it more flexible about how to produce the final results.

Additional context
------------------

sighingnow:
What do these changes do?
-------------------------

Add `accumulate=False/True` option to `vineyard.io.open()` to read CSV/Parquet/ORC directly to global dataframes.

Related issue number
--------------------

Fixes #1136


dashanji (approved):
Thanks for the fast and high-quality work!

PR #1141, opened 2023-01-10
Add container name as a option to kube_ssh.sh

dashanji:
What do these changes do?
-------------------------

As titled.

### DIFF
--- a/modules/io/python/drivers/io/kube_ssh.sh
+++ b/modules/io/python/drivers/io/kube_ssh.sh
@@ -3,32 +3,21 @@
 POD_NAME=$1
 shift
 
+KUBE_NS="default"
+KUBE_POD_NAME=""
+KUBE_CONTAINER_NAME=""
 if [[ "$POD_NAME" =~ ":" ]]; then
     KUBE_NS=${POD_NAME%%:*}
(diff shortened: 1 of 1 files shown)


modules/io/python/drivers/io/kube_ssh.sh
```diff
@@ -5,16 +5,28 @@ shift
...
+        kubectl -n ${KUBE_NS} exec ${KUBE_POD_NAME} -c ${KUBE_CONTAINER_NAME} -- /bin/bash -lc "
+            cat /etc/hosts > /dev/null || true && \
+            source /etc/profile || true && \
+            source /etc/bash.bashrc || true && \
+            source ~/.profile || true && \
+            source ~/.bashrc || true && \
+            shopt -s huponexit 2>/dev/null || true && \
+            $*"
+    else
+        kubectl -n ${KUBE_NS} exec ${KUBE_POD_CONTAINER_NAME} -c engine -- /bin/bash -lc "
```
sighingnow, line 21: Remove the default `engine` here.

We could just rely on the `default-container` setting in Pod's YAML.

dashanji: Done.

PR #1158, opened 2023-01-17
Drop the intermediate stream when ensure it has finished 

### LINKED ISSUE #1150: Delete the useless globalobjects and localobjects CR when loading data into vineyard cluster on kubernetes
Describe your problem
---------------------

When using the vineyard cluster on kubernetes and loading data into vineyard, we find some extra CRs like `vineyard::ParallelStream` and `vineyard::StreamCollection` will be created and it will confuse users to find the real data.

Expected
---------

We want these useless CRs will not to be created.

sighingnow:
What do these changes do?
-------------------------

- Add a `dropstream` method to vineyardd and client
- Add a finalizer in stream.py when finished consuming.

Related issue number
--------------------

Fixes #1150

### DIFF
--- a/.github/workflows/build-compatibility.yml
+++ b/.github/workflows/build-compatibility.yml
@@ -204,4 +204,6 @@ jobs:
           # build & install vineyard for python
 
           python3 setup.py bdist_wheel
-          sudo pip3 install dist/*.whl -U
+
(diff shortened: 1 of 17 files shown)


dashanji (approved):
Thanks.

PR #1160, opened 2023-01-17
Bisect rather than hashmap inside vertex map

sighingnow:
What do these changes do?
-------------------------

as titled.

Related issue number
--------------------

N/A


sighingnow:
Closing as archived.

PR #1165, opened 2023-01-31
Support configuring the operatiom image in the vineyardd CR and update relevant doc

dashanji:
What do these changes do?
-------------------------

* Support configuring the operation image in the vineyardd CR.
* Update the relevant doc.
* Generate the corresponding vineyard operator chart.

### DIFF
--- a/k8s/apis/k8s/v1alpha1/vineyardd_types.go
+++ b/k8s/apis/k8s/v1alpha1/vineyardd_types.go
@@ -167,9 +167,9 @@ type VineyardContainerConfig struct {
 	Env []corev1.EnvVar `json:"env,omitempty"`
 }
 
-// OperationImageConfig holds all image configuration about operations(backup, recover,
+// PluginImageConfig holds all image configuration about pluggable drivers(backup, recover,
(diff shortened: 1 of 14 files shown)


k8s/apis/k8s/v1alpha1/vineyardd_types.go
```diff
@@ -167,6 +167,35 @@ type VineyardContainerConfig struct {
 	Env []corev1.EnvVar `json:"env,omitempty"`
 }
 
+// OperationImageConfig holds all image configuration about operations(backup, recover,
+// local assembly, distributed assembly, repartition)
+type OperationImageConfig struct {
```
sighingnow, line 172: `PluginConfig` or `DriverConfig` would be a better name, as in our design we think they are pluggable drivers.

sighingnow (approved):

k8s/apis/k8s/v1alpha1/vineyardd_types.go
dashanji: Done and `Artifact Hub repository` updated too.

PR #1173, opened 2023-02-02
Replace the etcd image with vineyardd image

dashanji:
What do these changes do?
-------------------------

As titled.


sighingnow (commented):
Don't revert the submodule changes.

dashanji:
> Don't revert the submodule changes.

Done

PR #1177, opened 2023-02-03
Use k3d to replace kind for faster CI and less disk usage

sighingnow:


dashanji (approved):
Great work!! Big big thanks!!

PR #1183, opened 2023-02-07
Add vineyardctl for managing vineyard operator and relavant compenents on kubernetes

dashanji:
What do these changes do?
-------------------------

- Add vineyardoperatorctl(CLI tool) for deploying cert-manager and vineyard 
operator.
- Add relavant e2e test.

Related https://github.com/v6d-io/v6d/issues/1152

PR #1187, opened 2023-02-13
Bump up several components including golang, k8s api and controller-runtime

dashanji:
Bump up several compenents:
* Bump up golang to 1.19
* Bump up k8s api to 0.26.1
* Bump up controller-runtime to 0.14.4

What do these changes do?
-------------------------

As titled.


sighingnow (commented):
The test suite file should be automatically generated, why delete?

sighingnow:
> The test suite file should be automatically generated, why delete?

See https://github.com/v6d-io/v6d/pull/1183/commits/c35c42f26f2dd9d5b25a2445cbbe8e44ec85d2e9 for how to fixes the compilation error after bump up dependencies.

dashanji:
> The test suite file should be automatically generated, why delete?

It's only generated when creating a new crd, and actually, I think it's useless as we never use it. WDYT?

sighingnow:
> > The test suite file should be automatically generated, why delete?
> 
> It's only generated when creating a new crd, and actually, I think it's useless as we never use it. WDYT?

It if would be generated again after this deletion, e.g., adding new crds, please keep that, otherwise it would be added back someday.

dashanji:
Okay, I understand it. Thanks for pointing out the answer to fix the compilation error.

sighingnow (approved):

sighingnow (approved):

PR #1213, opened 2023-02-27
Add inject subcommand for sidecar injection and  {create/delete}operation subcommand

dashanji:
What do these changes do?
-------------------------

* Support reading the long configuration such as pv and pvc from stdin.
* Support injecting vineyard sidecar to a workload.
* Add the {create/delete}operation subcommand.

Related issue number
--------------------

Fixes #issue number


k8s/cmd/commands/sidecar/inject.go
```diff
@@ -0,0 +1,313 @@
...
+	// add labels to the workload for vineyard rpc service
+	labels[DefaultSidecarLabelName] = DefaultSidecarLabelValue
+
+	containers = append(containers, sidecarObj.Object["spec"].(map[string]interface{})["containers"].([]interface{})...)
+	volumes = append(volumes, sidecarObj.Object["spec"].(map[string]interface{})["volumes"].([]interface{})...)
+	templateSpec["containers"] = containers
+	templateSpec["volumes"] = volumes
+
+	return nil
+}
```
sighingnow, line 305: Could you please take a try to improve the code reuse? Looks quite same with https://github.com/v6d-io/v6d/blob/dev%2Fkubernetes/k8s/pkg/webhook/sidecar/sidecar_webhook.go#L190

Why it cannot be deserailzed as and object (rather than plain yaml object) then reuse some existing logic?

k8s/cmd/commands/util/file.go
```diff
@@ -0,0 +1,50 @@
...
+
+	fileSize := fileInfo.Size()
+	buffer := make([]byte, fileSize)
+
+	_, err = file.Read(buffer)
+	if err != nil {
+		return "", err
+	}
+
+	return string(buffer), nil
```
sighingnow, line 49: Why not

```
if content, err := iouti.ReadFile(path); err != nil {
   return "", err
} else {
   return string(content), nil
}
```

Don't repeat your self...

k8s/cmd/commands/sidecar/inject.go
```diff
@@ -0,0 +1,313 @@
...
+
+			yaml, err := util.ConvertToYaml(string(ss))
+			if err != nil {
+				return manifests, fmt.Errorf("failed to convert the unstructuredObj to yaml: %w", err)
+			}
+			yamls = append(yamls, yaml)
+		}
+	}
+
+	return yamls, nil
```
sighingnow, line 211: I cannot remember in which file I have meet this block, but it looks quite familiar. Maybe you could take a double-check to find if it can reuse some existing functions?

k8s/cmd/commands/util/parse.go
```diff
@@ -16,16 +16,72 @@ limitations under the License.
...
+	}
+
+	// Marshal the map into a YAML string
+	y, err := yaml.Marshal(m)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal to yaml: %v", err)
+	}
+
+	return string(y), nil
+}
```
sighingnow, line 65: Try https://pkg.go.dev/github.com/ghodss/yaml#section-readme

k8s/cmd/commands/util/parse.go
```diff
@@ -16,16 +16,72 @@ limitations under the License.
...
+			return "", fmt.Errorf("failed to read from stdin: %v", err)
+		}
+		j, err := ConvertToJson(string(r))
+		if err != nil {
+			return "", fmt.Errorf("failed to convert to json: %v", err)
+		}
+		return j, nil
+	}
+	return "", nil
+}
```
sighingnow, line 83: Don't mix unrelated things. You could have a function that only do the read. Otherwise you will need to copy this part to handle reading YAMLs from stdin someday later.

k8s/cmd/commands/flags/operation_flags.go
```diff
@@ -0,0 +1,43 @@
...
+
+func ApplyOperationName(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&OperationName, "name", "", "", "the name of operation")
+}
+
+func ApplyOperationOpts(cmd *cobra.Command) {
+	ApplyOperationName(cmd)
+	cmd.Flags().StringVarP(&OperationOpts.Name, "kind", "", "", "the kind of operation, including \"assembly\" and \"repartition\"")
+	cmd.Flags().StringVarP(&OperationOpts.Type, "type", "", "",
+		"the type of operation: for assembly, it can be \"local\" or \"distributed\"; for repartition, it can be \"dask\"")
```
sighingnow, line 39: Use single quote `'` for identifiers in docstring to avoid the escape characters.

k8s/cmd/commands/flags/operation_flags.go
```diff
@@ -0,0 +1,43 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package flags
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/v6d-io/v6d/k8s/apis/k8s/v1alpha1"
```
sighingnow, line 20: Grouping those imports would be better, see changes in my pull request.

sighingnow (approved):

dashanji:
All comments addressed and rebased, PLTA, thanks! @sighingnow

k8s/cmd/commands/create/create_backup.go
```diff
@@ -16,6 +16,8 @@ limitations under the License.
 package create
 
 import (
+	"log"
```
sighingnow, line 19: Use your `util.ErrLogger`. Mixing different logger in the same program is generally a bad practice.

k8s/cmd/commands/delete/delete_operation.go
```diff
@@ -0,0 +1,67 @@
...
+		if err := client.Get(context.Background(),
+			types.NamespacedName{Name: flags.OperationName,
+				Namespace: flags.GetDefaultVineyardNamespace()},
+			operation); err != nil && !apierrors.IsNotFound(err) {
+			util.ErrLogger.Fatal("failed to get operation: ", err)
+		}
+
+		if err := client.Delete(context.Background(), operation); err != nil {
+			util.ErrLogger.Fatal("failed to delete operation: ", err)
+		}
```
sighingnow, line 55: Use `util.Delete`.

k8s/cmd/commands/sidecar/inject.go
```diff
@@ -0,0 +1,311 @@
...
+		},
+		Spec: *opts,
+	}
+	return sidecar, nil
+}
+
+// InjectSidecarConfig injects the sidecar config into the workload
+func InjectSidecarConfig(sidecar *v1alpha1.Sidecar, workloadObj,
+	sidecarObj *unstructured.Unstructured) error {
+	errCollector := util.ErrorCollector{}
```
sighingnow, line 236: Consider use https://pkg.go.dev/go.uber.org/multierr.

k8s/cmd/commands/util/error.go
```diff
@@ -0,0 +1,43 @@
...
+	errStr := ""
+	if len(e.errlists) == 0 {
+		return nil
+	}
+
+	for _, e := range e.errlists {
+		errStr += e.Error() + ";"
+	}
+	return errors.New(errStr)
+}
```
sighingnow, line 43: See also https://pkg.go.dev/go.uber.org/multierr

k8s/cmd/commands/create/create_backup.go
dashanji: My bad. Thanks for pointing that out.

sighingnow:
Ping me once comments above get addressed.

dashanji:
Addressed.

PR #1262, opened 2023-03-21
fix data missing

### LINKED ISSUE #1261: data missing when loading csv
Describe your problem
---------------------

consider 9 node attending to load a 18 byte length csv file : 
id
91
292
293
294

header: id\n  (3byte)
data: 91\n292\n293\n294\n (15byte)

the code is 
```
def read_byte_blocks(
    fp,
    proc_num,
    proc_index,
    index,
    header_row,
    offset,
    chunk_size,
    read_block_delimiter,
    writer,
):
    try:
        total_size = fp.size()
    except TypeError:
        total_size = fp.size
    part_size = (total_size - offset) // proc_num
    begin = part_size * proc_index + offset
    end = min(begin + part_size, total_size)

    # See Note [Semantic of read_block with delimiter].
    if index == 0 and proc_index == 0:
        begin -= int(header_row) 
```
part_size = (18 - 3) // 9 = 1
and 9th node 
begin = 1 * (9 - 1) + 3 = 11
end = min(11 + 1, 18) = 12
and then the last line of csv is missing

ther is two way:

1. floor and function min are conflict, maybe use ceil to get part_size.
2. use `end = total_size if proc_index == proc_num - 1 else begin + part_size` to process the last proc

wuyueandrew:
What do these changes do?
-------------------------

assign total_size to the last proc

Related issue number
--------------------

Fixes #1261

### DIFF
--- a/modules/io/python/drivers/io/adaptors/read_bytes.py
+++ b/modules/io/python/drivers/io/adaptors/read_bytes.py
@@ -73,7 +73,7 @@ def read_byte_blocks(
         total_size = fp.size
     part_size = (total_size - offset) // proc_num
     begin = part_size * proc_index + offset
-    end = min(begin + part_size, total_size)
+    end = total_size if proc_index == proc_num - 1 else begin + part_size
 
     # See Note [Semantic of read_block with delimiter].
     if index == 0 and proc_index == 0:


sighingnow (approved):
LGTM. Nice catch!

PR #1286, opened 2023-04-03
Redefine the rbac roles of vineyard operator and update the helm chart

### LINKED ISSUE #1280: Redefine the RBAC roles of the Vineyard Operator
Describe your problem
---------------------

Thanks for Nanzi Yang's report via email. The following is the main content:
```
Dear vineyard maintainers:
I am Nanzi Yang, and I find a potential risk in the vineyard that can
be leveraged to make a cluster-level privilege escalation.

Detailed analysis:
The vineyard has two deployments called
"vineyard-operator-controller-manager" and "vineyardd-sample", which
run on worker nodes randomly. The service account of
controller-manager is "vineyard-operator-manager", and the service
account of vineyard-simple is "default". For these two service
accounts, both have a cluster role called
vineyard-operator-manager-role via different cluster role bindings.
The cluster role has "*" verbs of
"clusterrolebindings.rbac.authorization.k8s.io" and
"clusterroles.rbac.authorization.k8s.io" resources. Thus, if a
malicious user controls one worker node which has anyone of these two
deployments:
1. He/she can create/modify ClusterRoleBindings to bind an existing
cluster role (i.e., cluster-admin) to his/her service account,
resulting in cluster-level privilege escalation.
2. He/she can create/modify his/her own ClusterRoles to bind
"high-privileged" permissions to his/her own service account,
resulting in cluster-level privilege escalation.

Mitigation Discussion:
Perhaps the vineyard can remove the "modify" verbs of
ClusterRoleBinding and RoleBinding? However, it needs a careful review
of the vineyard's source code without disrupting its functionalities.

A few questions:
1. Is it a real issue in the vineyard?
2. If it's a real issue, can the vineyard mitigate the risk following
my suggestions?
Looking forward to your reply.
Regards,
Nanzi Yang

By the way, I tried to send this mail to security@v6d.io, however, it
said: "Mailbox size limit exceeded". So I find the maintainer list and
report this issue to you maintainers:)
```

As Nanzi Yang said, some RBAC roles have too much permission via https://github.com/v6d-io/v6d/blob/fb09f6465298172c7b054d5df4415d44e918c3a5/k8s/config/rbac/role.yaml#L299-L305.

The issue is to check all RBAC roles of vineyard operator for over request and update them to the proper role.

dashanji:
What do these changes do?
-------------------------

Redefine the over-request RBAC roles of vineyard operator.

Related issue number
--------------------

Fixes #1280


sighingnow:
Can those jobs created by vineyard-operator inherits the same cluster role with vineyard-operator itself?

Looks wired that we need to maintain those different cluster roles.

dashanji:
> Can those jobs **created by vineyard-operator** inherits the same cluster role with vineyard-operator itself?
> 
> Looks wired that we need to maintain those different cluster roles.

I'm sorry we can't. We assume the vineyard operator is trusted, and the jobs are created by users. If they have the same permission, the user can do anything in the role of vineyard operator. I think we need to set different permission for different tasks to achieve fine-grained permission control.

Also, we can update the `clusterrole/clusterrolebinding` to `role/rolebinding` so that all vineyard components will only be deployed in the same namespace and only have the permission to control the resources under the namespace such as  `vineyard-system`. WDUT? @sighingnow

sighingnow:
For "jobs created by vineyard-operator" I mean jobs like repartition, etc., created by controller in vineyard operator. Not user-defined jobs in their workflow.

> Also, we can update the clusterrole/clusterrolebinding to role/rolebinding so that all vineyard components will only be deployed in the same namespace and only have the permission to control the resources under the namespace such as vineyard-system.

OK.

Please double all occurances about "clusterroles" and "rolebindings" to make sure we don't overclaim.

dashanji:
Ok, I will update it later.

dashanji:
As discussed before, all vineyard jobs will share the default service account of vineyard operator. Also, minimize the permissions for roles in `mars examples`.

sighingnow (approved):
LGTM.

PR #1305, opened 2023-04-13
Use abstract socket address as the IPC socket

### LINKED ISSUE #1304: Use abstract socket on Linux to avoid requiring the physical UNIX-domain socket file
Describe your problem
---------------------

It is very inconvenient to have a socket file and mount the hostPath volume to pods on Kubernetes as well as using Docker.

Additional context
------------------

See also: https://man7.org/linux/man-pages/man7/unix.7.html

sighingnow:
Related issue number
--------------------

Fixes #1304


sighingnow:
Closing as it won't work on Kubernetes, as abstract socket is subject to `CLONE_NEWNET` however non-abstract UNIX-domain socket only relies on the file visibility.

PR #1311, opened 2023-04-13
Delete CRDs when deleting objects from vineyard

### LINKED ISSUE #1174: Dropping the intermediate stream didn't delete globalobjects and localobjects on kubernetes
Describe your problem
---------------------
Based on https://github.com/v6d-io/v6d/issues/1150.

The vineyard::ParallelStream globalobjects still exist but the metadata is missed. It looks like we need to add some judgments before vineyardd creates CR.
image

sighingnow:
What do these changes do?
-------------------------

The CRDs should be dropped from Kubernetes when we deleting the objects from vineyard. 

Related issue number
--------------------

- Fixes #1174
- See also: https://github.com/v6d-io/v6d/issues/1150#issuecomment-1413187626


dashanji (approved):
Thanks!!!

PR #1375, opened 2023-05-12
Improve the performance of `initDestFidList()` when there are many fragments

songqing:
What do these changes do?
-------------------------

Optimize the func initDestFidList(),  assuming edge num is n, fragment num is m, the original algo's time complexity is O(nlogm), the optimized is O(n)


sighingnow (approved):
LGTM.

PR #1402, opened 2023-05-29
Replace the limits with the objectIDs in the backup and recover of vineyard cluster.

### LINKED ISSUE #1322: Support applying raw backup and recover job for vineyard cluster
Describe your problem
---------------------

Although we have the API `vineyardctl create backup` and `vineyardctl create recover` to create the related job to backup objects in vineyard or recover from a backup, they are based on the fact that both the vineyard operator and the CRD have been installed.
It would be nice for users to create backup job and recover job directly via vineyardctl without deploying the operator and CRD.

We only need to consider one case: vineyard as deployment.

dashanji:
What do these changes do?
-------------------------

As titled. The `objectIDs` is more meaningful for users.

Related issue number
--------------------

Fixes #1322

### DIFF
--- a/k8s/controllers/k8s/backup_controller.go
+++ b/k8s/controllers/k8s/backup_controller.go
@@ -54,8 +54,8 @@ type FailoverConfig struct {
 
 // BackupConfig holds all configuration about backup
 type BackupConfig struct {
-	Name  string
-	Limit string
+	Name      string
+	ObjectIDs string
 	FailoverConfig
 }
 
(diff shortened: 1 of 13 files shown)


k8s/controllers/k8s/backup_controller.go
```diff
@@ -54,8 +54,8 @@ type FailoverConfig struct {
 
 // BackupConfig holds all configuration about backup
 type BackupConfig struct {
-	Name  string
-	Limit string
+	Name      string
+	ObjectIDs string
```
sighingnow, line 58: Use list of string if you would like to pass multiple value.

dashanji: Okay. That's good.

dashanji: Actually, I think it's better to make this consistent between vineyardctl and CRD to reduce the type convert. WDUT? @sighingnow 
https://github.com/v6d-io/v6d/pull/1402/files#diff-08fb2efc269a72bc81aedac5426c3b1cec120baae20021fa81beb33fcf0e9d68R52-R54

sighingnow: See https://pkg.go.dev/github.com/spf13/pflag#StringSlice

cobra support list arguments as well (via pflags). Don't use `,` or `;` to reinvent your own list stringify/encoding schemas.

dashanji: Thanks, make sense to me.

sighingnow (approved):

Issue #1373, opened 2023-05-12
Improve the memory usage of fragment (CSR) using delta encoding and varint


sighingnow:
Additional context
------------------

See also: https://github.com/alibaba/GraphScope/issues/2628

PR #1403, opened 2023-05-29
Batch encoding/decoding (with SIMD) of varint CSR

sighingnow:
What do these changes do?
-------------------------

- add back the offset array for correct degree
- use v8dec/nec for varint encoding
- implements batch decoding in CompactNbr

Related issue number
--------------------

Part of #1373


vegetableysm:
🐮

PR #1404, opened 2023-05-30
Improve GetInnerVertex() by querying local fragment only

songqing:
What do these changes do?
-------------------------

The current implementation of GetInnerVertex() is to query all the fragment, it's not needed as we only need the inner vertex


sighingnow (approved):

sighingnow:
Merged. Nice catch!

PR #1405, opened 2023-05-30
Fix a bug to avoid the panic 'interger divide by zero' during scheduling

dashanji:
What do these changes do?
-------------------------

As titled.

Related issue number
--------------------

Fixes some errors during e2e test such as https://github.com/v6d-io/v6d/actions/runs/5118226798/jobs/9202046172

### DIFF
--- a/k8s/pkg/schedulers/scheduling_strategy.go
+++ b/k8s/pkg/schedulers/scheduling_strategy.go
@@ -55,7 +55,7 @@ func NewRoundRobinStrategy(nodes []string) *RoundRobinStrategy {
 func (r *RoundRobinStrategy) Compute(rank int) (string, error) {
 	l := len(r.nodes)
 	if l == 0 {
-		return "", nil
+		return "", errors.Errorf("No nodes found")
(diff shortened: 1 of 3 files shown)


k8s/pkg/schedulers/scheduling_strategy.go
```diff
@@ -54,6 +54,9 @@ func NewRoundRobinStrategy(nodes []string) *RoundRobinStrategy {
 // Compute returns the node by the given rank.
 func (r *RoundRobinStrategy) Compute(rank int) (string, error) {
 	l := len(r.nodes)
+	if l == 0 {
+		return "", nil
+	}
```
sighingnow, line 59: Should we return an error if `nodes` is empty?

Return nil (success) with a invalid result `""` looks quite strange.

dashanji: Yeah, updated. Returning an error message is useful for debugging.

sighingnow (approved):

PR #1429, opened 2023-06-21
Fix ConcatenateTables()

songqing:
What do these changes do?
-------------------------

When calling ConcatenateTables(), the process will core as the out_tables[0] is not initialized.


sighingnow (approved):
Nice catch, Thanks for fixing it!

PR #1434, opened 2023-06-27
Fix bug: Inaccurate error message

vegetableysm:
What do these changes do?
-------------------------

When the path does not exist, concatenate the missing path information after the absolute path.

Related issue number
--------------------

Fixes [#2863](https://github.com/alibaba/GraphScope/issues/2863) (GraphScope)

### DIFF
--- a/modules/io/io/io_factory.cc
+++ b/modules/io/io/io_factory.cc
@@ -80,12 +80,22 @@ std::unique_ptr<IIOAdaptor> IOFactory::CreateIOAdaptor(
     auto s = uri.Parse(encoded_location);
     if (!s.ok()) {  // Assume it's a local file
       // defaulting to local file system: resolve to abs path first
-      char resolved_path[PATH_MAX];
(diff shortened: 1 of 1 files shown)


sighingnow (changes requested):
Actually, if you have tried some experiment code, you would find the (less-accurate) error message is thrown by apache-arrow from https://github.com/v6d-io/v6d/blob/main/modules/io/io/local_io_adaptor.cc#L157-L164.

So, I suggest some like

```c++
      auto r = fs_->OpenOutputStream(location_);
      if (r.ok()) {
        ofp_ = std::move(r).ValueOrDie();
      } else {
        return Status::Wrap(
            ArrowError(r.status()),
            "Failed to open file '" + location_ + "' with write mode");
      }
```

to serve as a wrapper to provide a more meaningful error message. That would be much simpler and enough to deliver a good error message to address the original issue.

sighingnow (approved):

PR #1438, opened 2023-06-28
Fix the sidecar can't work as expected and replace the kind version with v0.19.0

dashanji:
What do these changes do?
-------------------------

* Fix the bug the sidecar template is not injected into the workload yaml.
* Fix the bug the workload yaml can't contains the special character like '/' during sidecar injection.
* Add the ownerReferences to the Sidecar CR so that all injected resources will be deleted after deleting the workload.
* Add a test case for testing the vineyard sidecar via kubectl.
* Replace the kind version with v0.19.0 rather than v0.20.0.

Related issue number
--------------------

Related to #1355


sighingnow:
Hi @dashanji,

Could you please

1. add a more detailed description of this pull request, and 
2. if needed, add a regression test cases to prevent such a bug happen again?

sighingnow (approved):

PR #1440, opened 2023-06-29
Parallelize the PrepareToRun implementation in ArrowFragment

sighingnow:


sighingnow:
See also: https://github.com/alibaba/GraphScope/pull/2949

PR #1448, opened 2023-07-03
Bump up the k8s version to v1.25.10 to fix the existing Kubelet bug

dashanji:
What do these changes do?
-------------------------

As titled. Refer the info https://github.com/advisories/GHSA-xc8m-28vv-4pjc


sighingnow (approved):

dashanji:
It's hard to revert the k8s to an old version as the minimal k8s version of dependencies is v1.25.0, so it's better to backward compatible.

sighingnow (approved):

dashanji:
Context: https://github.com/kubernetes/kubernetes/issues/109164

PR #1483, opened 2023-07-21
Fixes the bug in vineyard docker image generation

sighingnow:


dashanji (approved):
Thanks for the quick fix!!

Issue #1503, opened 2023-08-04
Replace tracking chunks via Object CRD with the golang API in the command `vineyardctl schedule workflow`


dashanji:
Describe your problem
---------------------

To achieve the `best-effort` scheduling without the heavy scheduler-plugin, we propose the `vineyardctl schedule workflow` command which can schedule the workload in the `best-effort` policy with the default kubernetes scheduler. However, the current design requires the `GlobalObject` and `LocalObject` installed on the kubernetes cluster. To reduce the dependencies for users, we could use the current golang API to track chunks for different workloads.

PR #1507, opened 2023-08-09
Implement cluster-info API of Golang SDK and add the vineyardctl get cluster-info subcommand 

dashanji:
What do these changes do?
-------------------------

* Add the get cluster-info API for the golang sdk.
* Add the get cluster-info subcommand for vineyardctl.
* Generate the relevant doc.

Related issue number
--------------------

Fixes part of #1503


dashanji:
The vineyardctl e2e test will always crash as we have set the wrong image https://github.com/v6d-io/v6d/blob/main/k8s/config/manager/kustomization.yaml#L7 accidentally.

PR #1508, opened 2023-08-09
Use the golang sdk for tracking chunks for best-effort scheduling and add the relevant e2e test

### LINKED ISSUE #1503: Replace tracking chunks via Object CRD with the golang API in the command `vineyardctl schedule workflow`

dashanji:
What do these changes do?
-------------------------

As titled.

Related issue number
--------------------

Fixes #1503

### DIFF
--- a/.github/workflows/vineyard-operator.yaml
+++ b/.github/workflows/vineyard-operator.yaml
@@ -64,6 +64,7 @@ jobs:
           - e2e-tests-failover
           - e2e-tests-repartition-dask
           - e2e-tests-schedule-workflow
+          - e2e-tests-schedule-workflow-without-crd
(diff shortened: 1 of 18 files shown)


dashanji:
Wait for the previous pr #1507 to merge.

PR #1526, opened 2023-08-16
Fixes a bug in persist when multiple objects are involved

### LINKED ISSUE #1525: vineyardctl can't list all metadatas of vineyard deployment
Describe your problem
---------------------

When we use the `vineyardctl ls metadatas` to get the metadata of a vineyard deployment, assume the deployment has three replicas, and each vineyard pod is deployed on a different node. The current design can't list all metadatas of all vineyard pods as it can only connect one vineyard pod and get the relevant metadatas. The issue is to improve the `vineyardctl ls metadatas` command so that it can list all metadatas.

sighingnow:
Related issue number

Fixes #1525

### DIFF
--- a/python/vineyard/deploy/tests/test_distributed.py
+++ b/python/vineyard/deploy/tests/test_distributed.py
@@ -30,6 +30,7 @@ import numpy as np
 import pytest
 
 import vineyard
+from vineyard import ObjectMeta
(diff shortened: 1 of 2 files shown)


dashanji (approved):
Thanks a lot.

PR #1527, opened 2023-08-16
Use the crd-ref-docs to generate the docs for CRDs API automatically.

### LINKED ISSUE #1522: Split the CRDs of vineyard operator doc into the API Reference and revise these parts to make it easier to read and understand
Describe your problem
---------------------

As titled.

dashanji:
What do these changes do?
-------------------------

Fixes #1522. Wait for the [upstream default markers](https://github.com/elastic/crd-ref-docs/pull/49) ready, then it's enough for generating the essential APIs for users.


k8s/config.yaml
```diff
@@ -0,0 +1,20 @@
+# The file is used to generate the CRD reference documentation.
+# Copy from https://github.com/elastic/crd-ref-docs/blob/master/config.yaml
+
+processor:
+  ignoreTypes:
+    - "(Elasticsearch|Kibana|ApmServer)List$"
+    - "(Elasticsearch|Kibana|ApmServer)Health$"
+    - "(Elasticsearch|Kibana|ApmServer|Reconciler)Status$"
+    - "ElasticsearchSettings$"
+    - "Associa(ted|tor|tionStatus|tionConf)$"
```
sighingnow, line 10: We don't have those types. Remove it.

sighingnow (approved):
Why there are so many blank lines in the generated markdown?

k8s/config.yaml
dashanji: Done

dashanji:
> Why there are so many blank lines in the generated markdown?

I think it doesn't matter, you can see a similar doc like https://github.com/elastic/cloud-on-k8s/blob/main/docs/reference/api-docs.asciidoc.

k8s/test/e2e/workflow-demo/job1.py
```diff
@@ -23,6 +23,7 @@
 client = vineyard.connect('/var/run/vineyard.sock')
 
 object = client.put(np.arange(10))
+client.persist(object)
```
sighingnow, line 26: Can be avoided now.

```suggestion
```

PR #1534, opened 2023-08-22
Export the e2e test log when releasing and keep the vineyardd image consistent with the built one

dashanji:
What do these changes do?
-------------------------

* Fix the e2e test log that can't be exported when releasing.
* Keep the vineyardd image consistent with the previously built one.


sighingnow (approved):

dashanji:
Sometimes the e2e test fails, wait for a deep exploration here. https://github.com/v6d-io/v6d/actions/runs/5933760922/job/16089643240?pr=1533

PR #1538, opened 2023-08-25
Fix the vineyardctl e2e test errors and update the log package of vineyard scheduler

dashanji:
What do these changes do?
-------------------------

We found the vineyardctl e2e test always failed when we committed the PR to the main branch, the pr is to debug and fix it

### DIFF
--- a/.github/workflows/vineyard-operator.yaml
+++ b/.github/workflows/vineyard-operator.yaml
@@ -281,7 +281,7 @@ jobs:
           WITH_STOAT: true
 
       - name: Setup tmate session
-        if: ${{ failure() }}
+        if: false
         uses: mxschmitt/action-tmate@v3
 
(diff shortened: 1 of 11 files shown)


.github/workflows/vineyard-operator.yaml
```diff
@@ -280,6 +280,10 @@ jobs:
           OUTPUT_DIR: ${{ github.workspace }}/${{ matrix.job }}-logs
           WITH_STOAT: true
 
+      - name: Setup tmate session
+        if: ${{ failure() }}
```
sighingnow, line 284: Use `if: false`, and enable it when you want to debug.

dashanji: Done.

sighingnow (approved):

PR #1556, opened 2023-09-06
Fix bugs of hive module and modify some configurations.

vegetableysm:
Fix jimfs will throw exception when rename a dir.
Fix VineyardInputFormat wil throw exception when select an empty table.

What do these changes do?
-------------------------

- Set default storage and file format.
- Fix jimfs will throw exception when rename a dir.
- Fix VineyardInputFormat wil throw exception when select an empty table.
- Update readme.

Related issue number
--------------------

Fixes #issue number

### DIFF
--- a/java/hive/src/main/java/io/v6d/hadoop/fs/FileSystem.java
+++ b/java/hive/src/main/java/io/v6d/hadoop/fs/FileSystem.java
@@ -401,11 +401,7 @@ public class FileSystem extends org.apache.hadoop.fs.FileSystem {
                                 .replaceAll("\n", "");
                 IPCClient client = Context.getClient();
                 client.putName(ObjectID.fromString(objectIDStr), dstNioFilePath.toString());
-                try {
-                    client.dropName(srcNioFilePath.toString());
-                } catch (Exception e) {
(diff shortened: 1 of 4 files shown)


java/hive/src/main/java/io/v6d/hadoop/fs/FileSystem.java
```diff
@@ -373,25 +373,39 @@ public boolean renameInternal(Path src, Path dst) throws IOException {
...
+                int len = channel.read(bytes);
+                String objectIDStr =
+                        new String(bytes.array(), 0, len, StandardCharsets.UTF_8)
+                                .replaceAll("\n", "");
+                IPCClient client = Context.getClient();
+                client.putName(ObjectID.fromString(objectIDStr), dstNioFilePath.toString());
+                try {
+                    client.dropName(srcNioFilePath.toString());
+                } catch (Exception e) {
+                    Context.println("Drop name error: " + e.getMessage());
```
sighingnow, line 407: On which condition the drop name here will fail?

vegetableysm: Done.

PR #1557, opened 2023-09-07
Add vineyard.csi.read/write API for the integration with kubernetes volume.

### LINKED ISSUE #1555: Python SDK: PUT/GET API for vineyard csi driver
Describe your problem
---------------------

The issue is part of https://github.com/v6d-io/v6d/issues/1528, and also as a follow-up of https://github.com/v6d-io/v6d/pull/1533.

During integration with kubeflow, we need to provide a simple PUT/GET API to replace the original IO API(such as read_csv/to_csv/read_parquet/to_parquet and so on). The design is as follows:

vineyard.read('path')

```python

client = vineyard.connect('path/vineyard.sock')
objid = client.get_name(path)
value = client.get(path)
client.close()
return value

```

vineyard.write(value, 'path')

```python

client = vineyard.connect('path/vineyard.sock')
client.put(value, persist=True, name=path)
client.close()

```

dashanji:
What do these changes do?
-------------------------

As titled.

Related issue number
--------------------

Fixes #1555


python/vineyard/__init__.py
```diff
@@ -398,3 +398,54 @@ def connect(*args, **kwargs):
...
+        >>> arr = vineyard.read('/a/b/c/d/f')
+        >>> arr
+        array([0, 1, 2, 3, 4, 5, 6, 7])
+
+    Returns:
+        A python object that return by the resolver, by resolving an vineyard object.
+    """
+    socket_path = os.path.join(path, "vineyard.sock")
+    client = connect(socket_path)
+    return client.get(name=path)
```
sighingnow, line 451: Add a module `vineyard/csi/__init__.py` and put those two functions here, as they do strong assumption about the value of directory (where to find the socket and what should be used as names).

And, do check if the socket exists, and throw errors like "seems that the given path  is not generated by the vineyard CSI driver", or something like that.

sighingnow (approved):

python/vineyard/core/tests/test_client.py
```diff
@@ -281,3 +282,14 @@ def start_requests(rs, state, ipc_socket):
...
+
+def test_vineyard_read_and_write(vineyard_ipc_sockets):
+    path = os.path.dirname(vineyard_ipc_sockets[0])
+    # create a soft link of vineyard socket
+    os.symlink(vineyard_ipc_sockets[0], os.path.join(path, "vineyard.sock"))
+
+    value = (1, 2, 3, 4)
+    vineyard.write(value, path)
+    result = vineyard.read(path)
+    assert result == value
```
sighingnow, line 295: These APIs not should be tested using the CSI driver, not here.

As without the CSI driver, we never expect user have such path where both name and ipc socket location is embeded.

sighingnow (changes requested):

python/vineyard/core/tests/test_client.py
dashanji: Make sense.

python/vineyard/__init__.py
dashanji: Get it.

dashanji:
@sighingnow Please take another look. Also,  could you please consider releasing a version once the PR is merged? This way, I can build a benchmark test for the CSI driver using the new APIs. Thank you!

dashanji:
I have tested it locally but get the following error.

```bash
[error] json: [json.exception.parse_error.109] parse error: array index '/opt/caoye/test' is not a number in 'meta_[json::json_pointer(kv.key)] = value', at /work/v6d/src/server/services/meta_service.cc:354
[error] Check failed: Invalid: Failed to insert to metadata: key = '/names/~1opt~1caoye~1test', value = '6535182625925864', reason: Metatree invalid: json: [json.exception.parse_error.109] parse error: array index '/opt/caoye/test' is not a number in 'meta_[json::json_pointer(kv.key)] = value', at /work/v6d/src/server/services/meta_service.cc:354 in "status"
[error] json: [json.exception.parse_error.109] parse error: array index '/opt/caoye/test' is not a number in 'meta_[json::json_pointer(kv.key)] = value', at /work/v6d/src/server/services/meta_service.cc:354
```

sighingnow (approved):

PR #1564, opened 2023-09-14
Support getting FID from GID

sighingnow:


modules/graph/fragment/arrow_fragment.vineyard-mod
```diff
@@ -290,6 +290,8 @@ class [[vineyard]] ArrowFragment
     return IsInnerVertex(u) ? fid_ : vid_parser_.GetFid(GetOuterVertexGid(u));
   }
 
+  fid_t GetFragId(const gid_t& u) const { return vid_parser_.GetFid(u); }
```
songqing, line 293: vid_t? gid_t is not defined in v6d, and it's uint32 in Mac

sighingnow: Fixed in #1566, thanks!

PR #1597, opened 2023-10-20
Fix test of incremental update

songqing:
What do these changes do?
-------------------------
The current implementation of incremental update seems only support "retain_oid=true", however, the test hasn't set the option, and the test will core dump.

Related issue number
--------------------

f1


sighingnow:
Thanks for catching!

sighingnow (approved):

PR #1604, opened 2023-10-25
[GraphAr] Align the vertices of fragment to vertex chunk size & Fix the shuffle array multi-thread bug & Bump up GraphAr 

acezen:
What do these changes do?
-------------------------
- Align the vertices of fragment to vertex chunk size by append nulls, so the multi-fragment writer out to gar format also correct
- Fix the gar loader shuffle oid array bug: remove the multi-thread implementation since that shuffle already use thread to send/recv, that would make message conflict
- Bump up GraphAr to support build with local installed arrow.

Related issue number
--------------------
TBF

Fixes #issue number

### DIFF
--- a/modules/graph/fragment/gar_fragment_builder.h
+++ b/modules/graph/fragment/gar_fragment_builder.h
@@ -53,7 +53,9 @@ class GARFragmentBuilder
                               std::shared_ptr<vertex_map_t> vm_ptr)
       : ArrowFragmentBaseBuilder<oid_t, vid_t, vertex_map_t>(client),
         client_(client),
-        vm_ptr_(vm_ptr) {}
+        vm_ptr_(vm_ptr) {
+    Base::set_compact_edges_(false);
+  }
 
   vineyard::Status Build(vineyard::Client& client) override;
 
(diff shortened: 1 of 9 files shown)


modules/graph/fragment/gar_fragment_builder.h
```diff
@@ -53,7 +53,9 @@ class GARFragmentBuilder
                               std::shared_ptr<vertex_map_t> vm_ptr)
       : ArrowFragmentBaseBuilder<oid_t, vid_t, vertex_map_t>(client),
         client_(client),
-        vm_ptr_(vm_ptr) {}
+        vm_ptr_(vm_ptr) {
+    Base::set_compact_edges_(false);
```
acezen, line 57: Brief for why change here：`gar_fragment_builder` uses this constructor to construct builder and wants the `compact_edges` to be set false, but sometimes the constructor will initialize the `compact_edges` to be true, that would make builder corrupt.

sighingnow: I suggest you add an assertion and throw exception like "compacting edges is not supported when loading from GraphAr" or something like that.

sighingnow (approved):

sighingnow (approved):

PR #1615, opened 2023-11-15
Support to use RPC to get object and improve the resolver and builder of python

dashanji:
What do these changes do?
-------------------------

With it, rpc_client and ipc_client can get/put the value as the same way.

### DIFF
--- a/python/vineyard/contrib/ml/tests/test_dali.py
+++ b/python/vineyard/contrib/ml/tests/test_dali.py
@@ -32,6 +32,10 @@ def vineyard_for_dali():
         yield
 
 
+def test_dali_tensor_with_rpc_client(vineyard_rpc_client):
+    test_dali_tensor(vineyard_rpc_client)
+
+
 def test_dali_tensor(vineyard_client):
     @dali.pipeline_def()
     def pipe():
--- a/python/vineyard/core/resolver.py
+++ b/python/vineyard/core/resolver.py
@@ -230,12 +230,11 @@ def get(
         object_id = client.get_name(name)
 
     # run resolver
-    obj = client.get_object(object_id, fetch=fetch)
-    meta = obj.meta
-    if not meta.islocal and not meta.isglobal:
-        raise ValueError(
-            "Not a local object: for remote object, you can only get its metadata"
-        )
+    if isinstance(client, RPCClient):
+        obj = client.get_object(object_id)
+    else:
+        obj = client.get_object(object_id, fetch=fetch)
+
     if resolver is None:
         resolver = get_current_resolvers()
     return resolver(obj, __vineyard_client=client, **kw)
--- a/python/vineyard/data/arrow.py
+++ b/python/vineyard/data/arrow.py
@@ -31,16 +31,16 @@ except ImportError:
 from vineyard._C import Blob
 from vineyard._C import IPCClient
 from vineyard._C import Object
+from vineyard._C import ObjectID
 from vineyard._C import ObjectMeta
+from vineyard._C import RemoteBlob
 from vineyard.core.builder import BuilderContext
 from vineyard.core.resolver import ResolverContext
 from vineyard.data.utils import build_buffer
 from vineyard.data.utils import normalize_dtype
 
 
-def buffer_builder(
-    client: IPCClient, buffer: Union[bytes, memoryview], builder: BuilderContext
-):
+def buffer_builder(client, buffer: Union[bytes, memoryview], builder: BuilderContext):
     if buffer is None:
         address = None
         size = 0
@@ -51,7 +51,16 @@ def buffer_builder(
 
 
 def as_arrow_buffer(blob: Blob):
-    buffer = blob.buffer
+    if isinstance(blob, Blob):
+        buffer = blob.buffer
+    else:
+        if (
+            blob.id == ObjectID("o8000000000000000")
+            or (int(blob.id) & int(ObjectID("o8000000000000000"))) == 0
+        ):
+            buffer = b''
+        else:
+            buffer = memoryview(blob)
     if buffer is None:
         return pa.py_buffer(bytearray())
     return pa.py_buffer(buffer)
--- a/python/vineyard/data/arrow.py
+++ b/python/vineyard/data/arrow.py
@@ -33,7 +33,6 @@ from vineyard._C import IPCClient
 from vineyard._C import Object
 from vineyard._C import ObjectID
 from vineyard._C import ObjectMeta
-from vineyard._C import RemoteBlob
 from vineyard.core.builder import BuilderContext
 from vineyard.core.resolver import ResolverContext
 from vineyard.data.utils import build_buffer
--- a/python/vineyard/data/tests/test_arrow.py
+++ b/python/vineyard/data/tests/test_arrow.py
@@ -66,13 +66,27 @@ def test_arrow_array(vineyard_client):
     assert vineyard_client.get(object_id).values.equals(nested_arr.values)
 
 
-def test_record_batch(vineyard_client):
+def test_arrow_array_with_rpc_client(vineyard_rpc_client):
+    test_arrow_array(vineyard_rpc_client)
+
+
+def test_record_batch_with_rpc_client(vineyard_rpc_client):
+    test_record_batch(vineyard_rpc_client)
+
+
+@pytest.mark.skip
+def build_record_batch():
     arrays = [
         pa.array([1, 2, 3, 4]),
         pa.array(['foo', 'bar', 'baz', None]),
         pa.array([True, None, False, True]),
     ]
     batch = pa.RecordBatch.from_arrays(arrays, ['f0', 'f1', 'f2'])
+    return batch
+
+
+def test_record_batch(vineyard_client):
+    batch = build_record_batch()
     _object_id = vineyard_client.put(batch)  # noqa: F841
     # processing tables that contains string is not roundtrip, as StringArray
     # will be transformed to LargeStringArray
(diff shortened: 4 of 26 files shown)


python/vineyard/contrib/ml/tests/test_dali.py
```diff
@@ -32,6 +32,10 @@ def vineyard_for_dali():
         yield
 
 
+def test_dali_tensor_with_rpc_client(vineyard_rpc_client):
+    test_dali_tensor(vineyard_rpc_client)
+
+
```
sighingnow, line 38: Use `pytest.mark.parametrize` (https://docs.pytest.org/en/7.1.x/how-to/parametrize.html) to avoid these copies, e.g.,

```
@pytest.mark.parametrize("vineyard_client", [vineyard_client, vineyard_rpc_client])
def test....
```

python/vineyard/core/resolver.py
```diff
@@ -230,12 +230,11 @@ def get(
...
-    obj = client.get_object(object_id, fetch=fetch)
-    meta = obj.meta
-    if not meta.islocal and not meta.isglobal:
-        raise ValueError(
-            "Not a local object: for remote object, you can only get its metadata"
-        )
+    if isinstance(client, RPCClient):
+        obj = client.get_object(object_id)
+    else:
+        obj = client.get_object(object_id, fetch=fetch)
```
sighingnow, line 236: Add two methods `IsIPC()` and `IsRPC()` to the `class ClientBase` in C++ (you can default the return value to `false` and override the corresponding one in `class Client` and `class RPCClient`, respectfully). And expose them to Python bindings as `is_ipc` and `is_rpc` (as `property_readonly`).

Then revise all those `isinstance(...)` checks to use the property.

python/vineyard/data/arrow.py
```diff
@@ -51,7 +50,16 @@ def buffer_builder(
...
+    if isinstance(blob, Blob):
+        buffer = blob.buffer
+    else:
+        if (
+            blob.id == ObjectID("o8000000000000000")
+            or (int(blob.id) & int(ObjectID("o8000000000000000"))) == 0
+        ):
+            buffer = b''
+        else:
+            buffer = memoryview(blob)
```
sighingnow, line 62: We do have `is_empty`, https://github.com/v6d-io/v6d/blob/main/python/core.cc#L461.

You could add the comparison with empty blob id there, rather than hard code in Python code.

And, return both `memoryview(b'')` to keep the same return type in the if-then-else branches.

python/vineyard/data/tests/test_arrow.py
```diff
@@ -66,21 +66,40 @@ def test_arrow_array(vineyard_client):
...
     #
     # assert batch.equals(vineyard_client.get(object_id))
 
 
-def test_table(vineyard_client):
+def test_table_with_rpc_client(vineyard_rpc_client):
+    test_table(vineyard_rpc_client)
+
+
+@pytest.mark.skip
```
sighingnow, line 101: When skipping a test case, remember to add the reason: https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=skip#pytest.mark.skip

python/vineyard/data/tests/test_arrow.py
```diff
@@ -66,21 +66,40 @@ def test_arrow_array(vineyard_client):
...
+def build_tabel():
```
sighingnow, line 102: `tabel` -> `table`, and, with pytest, functions that don't starts with `test_` are not test cases and don't need the skip markers.

python/vineyard/data/tests/test_dataframe.py
```diff
@@ -165,3 +211,56 @@ def test_dataframe_reusing(vineyard_client):
...
+                [
+                    pd.Timestamp("2012-05-01"),
+                    pd.Timestamp("2012-05-02"),
+                    pd.Timestamp("2012-05-03"),
+                ]
+            )
+        ),
+    ],
+)
+def test_with_ipc_and_rpc(value, vineyard_client, vineyard_rpc_client):
```
sighingnow, line 258: Test what with ipc and rpc? Could you elaborate the function name to make it clear?

sighingnow: If you want to run the test case with both IPC client and rpc client, use `pytest.mark.parametrize`.

python/vineyard/data/utils.py
```diff
@@ -144,15 +147,30 @@ def ensure_ipc_client(client, error_message=None):
...
-    ensure_ipc_client(
-        client,
-        "Vineyard RPC client cannot be used to create local blobs, "
-        "try using an IPC client or `rpc_client.create_remote_blob()`",
-    )
+    if isinstance(client, RPCClient):
+        # copy the address with size to a local payloads
+        if size == 0 or address is None:
+            meta = ObjectMeta()
+            meta.id = ObjectID("o8000000000000000")
```
sighingnow, line 159: Don't hard code the empty blob in Python.

Like `client.create_empty_blob`, you could add such a method to RPC client as well, if needed.

src/client/client_base.cc
```diff
@@ -72,6 +72,11 @@ Status ClientBase::CreateData(const json& tree, ObjectID& id,
 }
 
 Status ClientBase::CreateMetaData(ObjectMeta& meta_data, ObjectID& id) {
+  if (this->instance_id_ == UnspecifiedInstanceID() - 1) {
```
sighingnow, line 75: What the "this->instance_id_ == UnspecifiedInstanceID() - 1" means?

src/client/ds/object_meta.cc
```diff
@@ -347,9 +363,15 @@ void ObjectMeta::SetMetaData(ClientBase* client, const json& meta) {
...
-      if (client_ == nullptr /* traverse to account blobs */ ||
-          tree["instance_id"].get<InstanceID>() == client_->instance_id()) {
+      if (client_ == nullptr) {
         VINEYARD_CHECK_OK(buffer_set_->EmplaceBuffer(member_id));
+      } else {
+        std::shared_ptr<struct InstanceStatus> instance_status = nullptr;
+        VINEYARD_CHECK_OK(client_->InstanceStatus(instance_status));
+        if (tree["instance_id"].get<InstanceID>() ==
+            instance_status->instance_id) {
+          VINEYARD_CHECK_OK(buffer_set_->EmplaceBuffer(member_id));
```
sighingnow, line 373: You would better cache the instance_status in RPCClient as a member, otherwise you would need call it for so many times when building/resolving metadata....

src/client/ds/object_meta.h
```diff
@@ -131,6 +131,11 @@ class ObjectMeta {
    */
   bool const IsLocal() const;
 
+  /**
+   * @brief Whether the object meta and rpc client is located on the same node.
+   */
+  bool const IsLocated() const;
```
sighingnow, line 137: Why we need it? Even the rpc client connects to the same node where the object live, it still a "remote" object and requires remote data access.

src/client/ds/remote_blob.cc
```diff
@@ -148,6 +149,48 @@ const std::shared_ptr<MutableBuffer>& RemoteBlobWriter::Buffer() const {
...
+          "but it "
+          "is nullptr: " +
+          ObjectIDToString(meta.GetId()));
+    }
+    this->size_ = this->buffer_->size();
+  } else {
+    throw std::runtime_error(
+        "RemoteBlob::Construct(): Invalid internal state: failed to construct "
+        "local "
+        "blob since payload is missing: " +
```
sighingnow, line 189: Can you merge the `"local "` and `"blob ..."` to the same line to make it better?

src/client/ds/remote_blob.cc
```diff
@@ -148,6 +149,48 @@ const std::shared_ptr<MutableBuffer>& RemoteBlobWriter::Buffer() const {
...
+        "is not located with rpc client");
+    return;
+  }
+
+  if (meta.GetBuffer(meta.GetId(), this->buffer_).ok()) {
+    if (this->buffer_ == nullptr) {
+      throw std::runtime_error(
+          "RemoteBlob::Construct(): Invalid internal state: remote blob found "
+          "but it "
+          "is nullptr: " +
```
sighingnow, line 181: Put `but it is nullptr` to the same line.

src/client/rpc_client.cc
```diff
@@ -180,11 +182,49 @@ Status RPCClient::GetObject(const ObjectID id,
   ObjectMeta meta;
   RETURN_ON_ERROR(this->GetMetaData(id, meta, true));
   RETURN_ON_ASSERT(!meta.MetaData().empty());
-  object = ObjectFactory::Create(meta.GetTypeName());
+  std::map<ObjectID, std::shared_ptr<Buffer>> buffers;
+  std::function<json&(json&)> travel = [&](json& meta_tree) -> json& {
```
sighingnow, line 186: traverse, not travel.

sighingnow (changes requested):

sighingnow:
Don't touch the submodule, if you are not actually changing it...

src/client/client_base.cc
dashanji: It means whether the client is a RPC client. https://github.com/v6d-io/v6d/blob/main/src/client/rpc_client.cc#L131

src/client/ds/object_meta.h
dashanji: Sorry about the confusing brief, actually it means whether the instance connected by rpc client is the same as meta's instance.

src/client/client_base.cc
sighingnow: Use the `IsIPC()` and `IsRPC` mentioned in another comment.

src/client/ds/object_meta.h
sighingnow: Add a method to the RPCClient:

```
bool IsFetchable(const ObjectMeta &object_meta) { ... }
```

PR #1622, opened 2023-11-27
Support to use RPC to get object and improve the resolver and builder of python

dashanji:
What do these changes do?
-------------------------

As titled.

Related issue number
--------------------

Replaces #1615

### DIFF
--- a/python/vineyard/contrib/ml/tests/test_dali.py
+++ b/python/vineyard/contrib/ml/tests/test_dali.py
@@ -20,7 +20,10 @@ import numpy as np
 
 import lazy_import
 import pytest
+import pytest_cases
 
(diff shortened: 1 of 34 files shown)


python/vineyard/contrib/ml/tests/test_dali.py
```diff
@@ -20,7 +20,10 @@
 
 import lazy_import
 import pytest
+from pytest_cases import parametrize
```
sighingnow, line 23: Use `import pytest_cases` and `@pytest_cases.parametrize(...)` anywhere you need it to avoid causing confusion with pytest's parametrize.

sighingnow (approved):

PR #1629, opened 2023-12-01
Add a wrapper to manage IPC and RPC clients and revise the statement about RPC client in documentation

dashanji:
What do these changes do?
-------------------------

- Update the doc for rpc client.
- Add the client wrapper to handle the IPCClient and RPCClient.


python/vineyard/core/client_manager.py
```diff
@@ -0,0 +1,90 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from vineyard._C import ObjectID
+from vineyard._C import _connect
+
+
+class ClientManager:
```
sighingnow, line 23: ClientManager -> Client.

You could rename the existing `Client` to `IPCClient` (only in python/client.cc).

sighingnow (changes requested):

python/vineyard/core/tests/test_client_on_distributed_vineyard.py
```diff
@@ -0,0 +1,81 @@
+#! /usr/bin/env python
```
sighingnow, line 1: I cannot see if there are any cases that can be covered by this test file but cannot be covered by existing test cases.

python/vineyard/core/resolver.py
```diff
@@ -229,21 +230,14 @@ def get(
...
-        raise RuntimeError('Unknown vineyard client type: %s' % type(client))
+    obj = client.get_object(object_id)
 
     if resolver is None:
         resolver = get_current_resolvers()
     return resolver(obj, __vineyard_client=client, **kw)
 
 
-setattr(IPCClient, 'get', get)
-setattr(RPCClient, 'get', get)
```
sighingnow, line 246: Keep the original bindings as well.

python/vineyard/core/client.py
```diff
@@ -0,0 +1,422 @@
...
+                    username=username,
+                    password=password,
+                )
+
+        if self._ipc_client is None and self._rpc_client is None:
+            raise ConnectionError(
+                "Failed to resolve IPC socket or RPC endpoint of vineyard server from "
+                "environment variables VINEYARD_IPC_SOCKET or VINEYARD_RPC_ENDPOINT."
+            )
+        return self
```
sighingnow, line 103: Please put contents in `connect()` to `__init__()` to make it more Pythonic.

sighingnow (changes requested):
Please address these "Either all return statements in a function should return an expression, or none of them should." warnings.

python/vineyard/core/tests/test_client_on_distributed_vineyard.py
dashanji: Actually, the main purpose of this test is to get the same data from different vineyard nodes.

python/vineyard/core/client.py
dashanji: Done.

python/vineyard/core/tests/test_client_on_distributed_vineyard.py
sighingnow: Move the test case to https://github.com/v6d-io/v6d/blob/main/python/vineyard/deploy/tests/test_distributed.py

PR #1646, opened 2023-12-13
Implement concurrent memcpy for building Python objects into vineyard

### LINKED ISSUE #1631: Poor performance
Hi, 

Under several benchmarks of putting data object to the shared store (from 1KB to several GBs), we observed that `vineyard` underperforms `ray` (`plasma`), for spending 2x-5x more time

With the data object size getting larger, the performance issue scales. Are there any specific reasons or sources of overhead?

sighingnow:
Remove the problematic `.buffer` property (as it cannot bind the lifetime of the underlying blob to the memoryview object) and add concurrent support for memcpy for faster object building.

Fixes #1631

### DIFF
--- a/python/core.cc
+++ b/python/core.cc
@@ -22,6 +22,7 @@ limitations under the License.
 #include "client/ds/object_meta.h"
 #include "client/ds/remote_blob.h"
 #include "client/rpc_client.h"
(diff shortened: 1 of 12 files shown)


dashanji (approved):
Thanks for claifying!

PR #1667, opened 2023-12-18
Add a const method for GetVertexMap

songqing:
What do these changes do?
-------------------------
We I want to use Arrow fragment's method GetVertexMap() in GAE apps, there will be a compile error, as the fragment is const, but the method is not const, so, add a new const method for GAE apps.
`void PEval(const fragment_t& frag, context_t& ctx,
             message_manager_t& messages) `

### DIFF
--- a/modules/graph/fragment/arrow_fragment.vineyard-mod
+++ b/modules/graph/fragment/arrow_fragment.vineyard-mod
@@ -580,7 +580,7 @@ class [[vineyard]] ArrowFragment
                            iodoffset_[v_label][e_label][offset + 1]);
   }
 
-const std::shared_ptr<vertex_map_t> GetVertexMap() const { return vm_ptr_; }
+  const std::shared_ptr<vertex_map_t> GetVertexMap() const { return vm_ptr_; }
 
   std::shared_ptr<vertex_map_t> GetVertexMap() { return vm_ptr_; }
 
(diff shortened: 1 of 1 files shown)


sighingnow (approved):
LGTM. Thanks for fixing.

modules/graph/fragment/arrow_fragment.vineyard-mod
```diff
@@ -580,6 +580,10 @@ class [[vineyard]] ArrowFragment
                            iodoffset_[v_label][e_label][offset + 1]);
   }
 
+  const std::shared_ptr<vertex_map_t> GetVertexMap() const {
+    return vm_ptr_;
+  }
```
sighingnow, line 585:
```suggestion
const std::shared_ptr<vertex_map_t> GetVertexMap() const { return vm_ptr_; }
```

modules/graph/fragment/arrow_fragment.vineyard-mod
```diff
@@ -580,6 +580,8 @@ class [[vineyard]] ArrowFragment
                            iodoffset_[v_label][e_label][offset + 1]);
   }
 
+const std::shared_ptr<vertex_map_t> GetVertexMap() const { return vm_ptr_; }
```
sighingnow, line 583:
```suggestion
  const std::shared_ptr<vertex_map_t> GetVertexMap() const { return vm_ptr_; }
```

PR #1688, opened 2023-12-20
Simplify the buffer builder for RPC client

### LINKED ISSUE #1664: Avoid excessive memory copy in remote blob builder
See also: https://github.com/v6d-io/v6d/blob/63dd6adb5473e002889485f4afecf47a9d0d1094/python/vineyard/data/utils.py#L144-L183

sighingnow:
Fixes #1664


dashanji (approved):
Thanks for fixing it.

Issue #1596, opened 2023-10-20
Integrate with fluid


dashanji:
Describe your problem
---------------------

**Motivation**
Use [Fluid](https://github.com/fluid-cloudnative/fluid) as the launcher of vineyard. 

**Detail**
We can leverage the [Thinruntime and Dataset mechanism](https://github.com/fluid-cloudnative/fluid/blob/master/docs/zh/samples/thinruntime.md) to mount the vineyard socket so that we don't need to deploy the Vineyard CSI Driver anymore.

sighingnow:
Closing as fluid PRs has been merged.

PR #1704, opened 2023-12-29
Add the mount vineyard socket dockerfile for fluid integration

dashanji:
What do these changes do?
-------------------------

Add the mount vineyard socket dockerfile and we need to push it to our official docker hub when merged.

Related issue number
--------------------

Fixes part of #1596

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,4 +1,5 @@
 REGISTRY 					:= ghcr.io/v6d-io/v6d
+RELEASE_REGISTRY			:= docker.io/vineyardcloudnative
 
 # can be: x86_64, aarch64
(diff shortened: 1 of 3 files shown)


sighingnow:
Seems this artifacts should be part of vineyard fluid runtime?

dashanji:
> Seems this artifacts should be part of vineyard fluid runtime?

I think it should be a part of vineyard just as `vineyard fuse`.

sighingnow:
It can only be used in the context of fluid ....

sighingnow:
Rename the dir and docker image name to `vineyard-fluid-mount`.

sighingnow (approved):

PR #1711, opened 2024-01-08
Avoid to hardcode the fuse dir and change the image from vineyard-fluid-mount to vineyard-fluid-fuse

dashanji:
What do these changes do?
-------------------------

As titled.


sighingnow (approved):

sighingnow (changes requested):
It is not a "fuse" driver....

dashanji:
> It is not a "fuse" driver....

Yeah, but it can be consistent with the other runtime in the fluid system.
https://github.com/fluid-cloudnative/fluid/pull/3649#discussion_r1444195408

PR #1735, opened 2024-01-26
Refactor code of radix split subtree and fix bug of radix split.

vegetableysm:
What do these changes do?
-------------------------

As tittled.

Related issue number
--------------------

Fixes #issue number


dashanji (approved):
@vegetableysm  Thanks a lot!!!

PR #1736, opened 2024-01-29
Check whether the ipc socket exist When parsing the config YAML.

dashanji:
What do these changes do?
-------------------------

When using something like `client = vineyard.connect(config=./vineyard.yaml)`, we need to check whether the IPC socket file exists.


dashanji:
The latest `pytest`(8.0.0) will block something of `pytest-cases`. 
https://github.com/smarie/python-pytest-cases/issues/330

sighingnow (approved):

PR #1760, opened 2024-02-19
Adapt k8s to v1.19.3.

### LINKED ISSUE #1651: Make Vineyard Operator compatible with a certain range of lower versions of k8s
Describe your problem
---------------------

### What is the problem:

Recently, when users were running some examples with a lower version of k8s (e,g. 1.23). Some components of vineyard operator can't work as expected.

### The behaviour that you expect to work:

It's better to make vineyard operator compatible with a certain range of lower versions of k8s (1.19~). A possible way is to 
 writer a wrapper, which can build the different images of vineyard operator for different k8s version and push the image as `vineyardcloudnative/vineyard-operator:(vineyard-version)-(k8s-version)` such as `vineyardcloutnative/vineyard-operator:0.19.1-k8s1.19`. Besides, we need to add a flag `k8s.version` to the helm chart.

dashanji:
What do these changes do?
-------------------------

- Adapt k8s to v1.19.3.
- ~~Replace the log package in `go/vineyard` with `k8s/pkg/log`.~~
- Use the log implementation under `k8s/pkg` to replace the log implementation under `go/vineyard`.

Related issue number
--------------------

Fixes #1651


sighingnow:
> Replace the log package in go/vineyard with k8s/pkg/log.

Don't do that. The package `go/vineyard` is expected to be used as a standalone package and requiring k8s components/packages in such a data structure package looks quite wired.

dashanji:
Ok, I will try to fix the version compatibility between `go/vineyard/log` and `k8s/pkg/log`.

dashanji:
Sorry for the previous confusing description and please check the updated one. Actually, the k8s@v1.19 depends on a low version of `github.com/go-logr/logr` but the current implementation of `go/vineyard` requires a high version. Thus, I implemented the log package in the `go/vineyard` and reused the logic in the `k8s/pkg/log`.

dashanji:
The vineyard scheduler is closely related to the k8s version. As the scheduler configurations of lower versions will be deprecated in higher versions of k8s, we can't support multiple versions (v1beta1, v1beta2, v1beta3, and v1) of scheduler configuration in the same codebase.

A possible way is to find the lowest version of scheduler configuration and make sure it can work in the latest k8s cluster.

PR #1762, opened 2024-02-19
go: Adapt to older version of logr (and zap)

sighingnow:


dashanji:
Close via has been implemented in https://github.com/v6d-io/v6d/pull/1760

PR #1763, opened 2024-02-20
Create a hard link of vineyard socket rather than mount its parent directory.

dashanji:
What do these changes do?
-------------------------

- Create a hard link of vineyard socket rather than mount its parent directory in `vineyard-fluid-fuse`.
- Change the vineyard configuration from `vineyard.yaml` to `config.yaml`.

Related issue number
--------------------

Related to https://github.com/fluid-cloudnative/fluid/pull/3713


sighingnow:
> Change the vineyard configuration from vineyard.yaml to config.yaml.

Too confusing. `vineyard-config.yaml` would be better.

dashanji:
Updated.

sighingnow (approved):

PR #1765, opened 2024-02-22
GraphAr: Upgrade GraphAr to v0.11.2 and revise the gar loader and writer

acezen:
What do these changes do?
-------------------------
- Upgrade GraphAr to v0.11.2 and adapt the `GARFragmentLoader` and `ArrowFragmentWriter` with latest GraphAr API
- Support generate the GraphInfo with schema of property graph.
- Support dumps and load `ArrowFrgment` with GraphAr format with local in different workers. (We only support write and load to a shared location like nfs or OSS before)

Related issue number
--------------------

Fixes #issue number


acezen:
cc/ @sighingnow

PR #1794, opened 2024-03-04
Change llm api return value to vineyard::Status.

vegetableysm:
What do these changes do?
-------------------------

Related issue number
--------------------

Fixes #1791

PR #1797, opened 2024-03-04
Improve the query API of llm cache and use vector<uint8_t> as payload object.

### LINKED ISSUE #1786: llm-cache: double check all usage of call-by-value of `std::vector<T>` and `std::shared_ptr<T>`
As these copy may become performance bottleneck.

e.g., I have noticed some usage of `std::vector<int> keys` as argument.

### LINKED ISSUE #1792: llm-cache: avoid copying during query kv caches
This is wrong. You should return the shared-memory's pointer instead.

image

### LINKED ISSUE #1795: llm-cache: use `vector<uint8_t>` as payload object and remove all assumption about double
image

dashanji:
What do these changes do?
-------------------------

- Improve the `query` API, users only input a token list and will get the kv_cache with the longest prefix.
- Use vector<uint8_t> as payload object.
- Replace the alias of KV_STATE_WITH_LAYER with std::map<int, std::pair<K_STATE, V_STATE>>.
- Rename the `Dimension` with `TensorBytes`.
- Use the references of `std::vector<T>` to avoid copying.
- Print the rax tree to a string for debugging.

Related issue number
--------------------

- Fixes #1786
- Fixes #1792
- Fixes #1795

### DIFF
--- a/modules/basic/ds/dataframe.cc
+++ b/modules/basic/ds/dataframe.cc
@@ -85,6 +85,12 @@ const std::shared_ptr<arrow::RecordBatch> DataFrame::AsBatch(bool copy) const {
     } else if (auto tensor =
                    std::dynamic_pointer_cast<Tensor<std::string>>(df_col)) {
       num_rows = tensor->shape()[0];
+    } else if (auto tensor =
+                   std::dynamic_pointer_cast<Tensor<uint8_t>>(df_col)) {
+      num_rows = tensor->shape()[0];
+    } else if (auto tensor =
+                   std::dynamic_pointer_cast<Tensor<int8_t>>(df_col)) {
+      num_rows = tensor->shape()[0];
     }
 
     std::vector<std::shared_ptr<arrow::Buffer>> buffer{
--- a/modules/llm-cache/ds/kv_state_cache.cc
+++ b/modules/llm-cache/ds/kv_state_cache.cc
@@ -45,7 +45,9 @@ void KVStateCache::Resolve() {
   // 1. construct the radix tree
   this->rootTree = RadixTree::Deserialize(
       base64_decode(this->meta_.GetKeyValue<std::string>("radix_tree")));
-  // raxShow(this->rootTree->GetRootTree());
+  if (VLOG_IS_ON(100)) {
+    VLOG(100) << raxShow(this->rootTree->GetRootTree());
+  }
 
   // 2. construct the kvStateCacheBlockBuilder list
   size_t numBlocks = this->meta_.GetKeyValue<size_t>("numBlocks");
@@ -57,24 +59,24 @@ void KVStateCache::Resolve() {
   }
 
   // 3. construct the member field
-  this->dimension = this->meta_.GetKeyValue<int>("dimension");
+  this->tensorBytes = this->meta_.GetKeyValue<int>("tensorBytes");
   this->version = this->meta_.GetKeyValue<uint64_t>("version");
   this->layer = this->meta_.GetKeyValue<int>("layer");
-  VLOG(100) << "construct the member field success, with dimension:"
-            << this->dimension << " version:" << this->version
+  VLOG(100) << "construct the member field success, with tensorBytes:"
+            << this->tensorBytes << " version:" << this->version
             << " layer:" << this->layer;
 }
 
 KVStateCache::~KVStateCache() {}
 
-KVStateCacheBuilder::KVStateCacheBuilder(Client& client, int dimension,
+KVStateCacheBuilder::KVStateCacheBuilder(Client& client, int tensorBytes,
                                          int cacheCapacity, int layer,
                                          int blockSize) {
-  this->dimension = dimension;
+  this->tensorBytes = tensorBytes;
   this->version = 0;
   this->layer = layer;
   KVStateCacheBlockBuilder* builder =
-      new KVStateCacheBlockBuilder(client, this->dimension, layer, blockSize);
+      new KVStateCacheBlockBuilder(client, this->tensorBytes, layer, blockSize);
 
   this->rootTree = std::make_shared<RadixTree>(cacheCapacity);
 
@@ -90,7 +92,7 @@ KVStateCacheBuilder::KVStateCacheBuilder(Client& client, int dimension,
 
 KVStateCacheBuilder::KVStateCacheBuilder(Client& client,
                                          std::shared_ptr<KVStateCache> cache) {
-  this->dimension = cache->GetDimension();
+  this->tensorBytes = cache->GetTensorBytes();
   this->version = cache->GetVersion();
   this->layer = cache->GetLayer();
   // 1. create block builder from block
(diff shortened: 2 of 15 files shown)


sighingnow (approved):

modules/basic/ds/dataframe.cc
```diff
@@ -85,6 +85,9 @@ const std::shared_ptr<arrow::RecordBatch> DataFrame::AsBatch(bool copy) const {
     } else if (auto tensor =
                    std::dynamic_pointer_cast<Tensor<std::string>>(df_col)) {
       num_rows = tensor->shape()[0];
+    } else if (auto tensor =
+                   std::dynamic_pointer_cast<Tensor<uint8_t>>(df_col)) {
+      num_rows = tensor->shape()[0];
```
sighingnow, line 90: Add condition `int8_t` as well.

sighingnow (changes requested):
`Client &client` could be a member of the `KVStateCacheBuilder` (as well as `KVStateCacheBlockBuilder`). Don't given every APIs a `Client &client` argument.

You can image that user may pass different client object to the same `KVStateCacheBuilder`.

sighingnow (changes requested):
Rename `K_STATE` and `V_STATE` to `LLMKV`, or `llm_kv_t`.

modules/llm-cache/ds/kv_state_cache_block.h
```diff
@@ -170,20 +167,21 @@ class KVStateCacheBlockBuilder : public ObjectBuilder {
...
   }
 
-  const std::shared_ptr<TensorBuilder<double>> GetValueStateBuilder(int layer) {
+  const std::shared_ptr<TensorBuilder<uint8_t>> GetValueStateBuilder(
+      int layer) {
     return valueStateTensorBuilderList[layer];
   }
 
-  const std::vector<std::shared_ptr<TensorBuilder<double>>>
+  const std::vector<std::shared_ptr<TensorBuilder<uint8_t>>>
```
sighingnow, line 179: I believe return a `const std::vector<T> &` (i.e., const reference of the vector) would be better for these getters to avoid copy.

modules/llm-cache/tests/kv_state_cache_benchmark_test.cc
```diff
@@ -27,7 +27,7 @@ limitations under the License.
 
 using namespace vineyard;  //  NOLINT(build/namespaces)
 
-#define DIMENSION 100
+#define TENSORBYTES 800
 #define CAPACITY 1000
 #define LAYER 64
 #define BLOCK_SIZE 100
```
sighingnow, line 33: Use `constexpr int` or `constexpr size_t` for constants in C++.

modules/llm-cache/tests/kv_state_cache_test.cc
```diff
@@ -24,7 +24,7 @@ limitations under the License.
 
 using namespace vineyard;  // NOLINT(build/namespaces)
 
-int dimension = 10;
+int tensorBytes = 80;
 int capacity = 20;
 int layer = 3;
 int block_size = 5;
```
sighingnow, line 30: Use `constexpr int` instead.

sighingnow (changes requested):

modules/basic/ds/dataframe.cc
dashanji: ok

modules/llm-cache/tests/kv_state_cache_test.cc
dashanji: Done.

modules/llm-cache/ds/kv_state_cache_block.h
dashanji: It's a big change if we use the template to replace `TensorBuilder<uint8_t>`, could we open another issue to address it?

sighingnow: Why a big change?

dashanji: It's under the basic class of llm kv cache, If we change the type, the loc will be large. If you don't think it's hard to review, we can implement it here.

modules/llm-cache/ds/kv_state_cache.cc
```diff
@@ -45,7 +45,7 @@ void KVStateCache::Resolve() {
   // 1. construct the radix tree
   this->rootTree = RadixTree::Deserialize(
       base64_decode(this->meta_.GetKeyValue<std::string>("radix_tree")));
-  // raxShow(this->rootTree->GetRootTree());
+  VLOG(100) << raxShow(this->rootTree->GetRootTree());
```
sighingnow, line 49: Use

```
if (V_LOG_IS_ON(100)) {
  VLOG(100) ...
}
```

to avoid expensive stringifying of the arguments when GLOG_v<100.

modules/llm-cache/ds/kv_state_cache_block.cc
```diff
@@ -164,23 +167,21 @@ bool KVStateCacheBlockBuilder::IsFull() {
...
-    double* keyData = keyStateTensorBuilderList[currentLayer]->data();
-    double* valueData = valueStateTensorBuilderList[currentLayer]->data();
-    memcpy(keyData + index * this->dimension, keyState.data,
-           this->dimension * sizeof(double));
-    memcpy(valueData + index * this->dimension, valueState.data,
-           this->dimension * sizeof(double));
+    LLMKV keyState = (kvState.find(currentLayer)->second).first;
+    LLMKV valueState = (kvState.find(currentLayer)->second).second;
+    VINEYARD_ASSERT(keyState.length == (size_t) this->tensorBytes);
+    VINEYARD_ASSERT(valueState.length == (size_t) this->tensorBytes);
```
sighingnow, line 177: Let `Update()` returns status and use `RETURN_ON_ERROR/RETUREN_ON_ASSERT`.

@vegetableysm remember to revise all usage of `VINEYARD_ASSERT` to see if it can be replaced by `RETUREN_ON_ASSERT`.

modules/llm-cache/radix-tree/radix-tree.cc
```diff
@@ -60,30 +60,30 @@ RadixTree::~RadixTree() {
...
 std::vector<std::shared_ptr<NodeData>> RadixTree::Split(
-    std::vector<int> tokens, std::shared_ptr<NodeData>& header) {
+    std::vector<int>& tokens, std::shared_ptr<NodeData>& header) {
   tokens.insert(tokens.begin(), INT32_MAX);
   return SplitInternal(tokens, header);
 }
 
 std::shared_ptr<NodeData> RadixTree::InsertInternal(
-    std::vector<int> tokens, std::shared_ptr<NodeData>& evictedNode) {
+    std::vector<int>& tokens, std::shared_ptr<NodeData>& evictedNode) {
```
sighingnow, line 86: Should be `const std::vector<int>& tokens` as it is the argument, not for return value.

sighingnow (changes requested):

modules/llm-cache/ds/kv_state_cache_block.cc
dashanji: Implemented in https://github.com/v6d-io/v6d/pull/1794/files#diff-58f18c2f80244c1a7accb540ac9d106b6f920e7679fee6591f6cc30d95018cadR168-R191

Issue #1802, opened 2024-03-06
Refactor radix tree to avoid memory copy.


vegetableysm:
Describe your problem
---------------------

Currently, the radix tree will insert a root node as the first "subtree". The node value is INT32_MAX. So when query a token list, it need to insert INT32_MAX to the begin of the token list. We need to refactor radix tree to avoid it. e.g.:

- Make compare of root node as a builtin function in raxLowWalk
or
- Use another field to represent the rooted tree.
or
- ...

PR #1805, opened 2024-03-07
Update the token list type from int* to vector<int> to avoid copying.

dashanji:
What do these changes do?
-------------------------

- Update the token list type from int* to vector<int> to avoid copying in the radix tree.
- Reduce the memory usage in the merge tree function.

Related issue number
--------------------

Fixes https://github.com/v6d-io/v6d/issues/1802


thirdparty/rax/radix.cc
```diff
@@ -48,6 +48,7 @@
 
 #include <iostream>
 #include "common/util/logging.h"
+#include "common/backtrace/backtrace.hpp"
```
sighingnow, line 51: ?

thirdparty/rax/radix.cc
```diff
@@ -2212,6 +2238,10 @@ std::string raxShow(rax *rax) {
     raxStr += "rax numnode:" + std::to_string(rax->numele) + "\n";
     raxStr += raxRecursiveShow(0,0,rax->head);
     raxStr += "\n";
+    // also show the backtrace info
+    std::stringstream ss;
+    backtrace_info::backtrace(ss, true);
+    raxStr += ss.str();
```
sighingnow, line 2244: Display back trace is costly, can you please wrap it into a `#ifndef _NDEBUG` block?

sighingnow (approved):

thirdparty/rax/radix.cc
dashanji: Okay.

vegetableysm (approved):

PR #1806, opened 2024-03-07
Remove block objects from cache object meta to reduce the meta size.

### LINKED ISSUE #1733: Only pull the cache blocks requested by the user
Describe your problem
---------------------

Avoid pull all the cache block from remote.
Do not add all block objects to cache object as members.

Additional context
------------------

vegetableysm:
What do these changes do?
-------------------------

As tittled.

Related issue number
--------------------

Fixes #1733


vegetableysm:
Before this optimize:
image
With kv_state_cache_benchmark_tes.cc(In https://github.com/v6d-io/v6d/pull/1816) and the config with:
```
constexpr int TENSORBYTES = 80;
constexpr int CAPACITY = 20000;
constexpr int LAYER = 64;
constexpr int BLOCK_SIZE = 100;
```
And the token list length is 1900;

So, the llm cache object contains 370 * 128 = 47360 members.

After this optimize, the cache object do not make the block as its member. So, the object with the largest number of members is cache block, which has 128 members

sighingnow (approved):

PR #1815, opened 2024-03-12
Fix llm memory leak and replace mutex_lock with unique_lock

### LINKED ISSUE #1813: There exists memory leak in llm cache.
Describe your problem
---------------------

When I run the test program multiple times and output the memory usage at the end of it, it got bigger and bigger each time.
```
client.InstanceStatus(status);
LOG(INFO) << status.memory_usage;
```

The first time
image

The second time 
image

There must exists memory leak.

### LINKED ISSUE #1814: Use lock_guard or unique_lock to manager locks in llm module.
Describe your problem
---------------------

As tittled.

vegetableysm:
What do these changes do?
-------------------------

- Fix memory leak of llm module.
- Use unique_lock to manager mutex locks in llm module.

Related issue number
--------------------

Fixes #1813 
Fixes #1814

### DIFF
--- a/modules/llm-cache/ds/kv_state_cache.cc
+++ b/modules/llm-cache/ds/kv_state_cache.cc
@@ -259,6 +259,9 @@ void KVStateCacheBuilder::Delete(std::shared_ptr<NodeData> evictedNodeData) {
   // delete (DataWrapper*) evictedNodeData->nodeData;
   if (evictedNodeData->cleanTreeData) {
     this->rootTree->ClearSubtreeData(treeData);
+    std::shared_ptr<Object> blockObject =
+        kvStateCacheBlockBuilder->_Seal(client);
+    client.DelData(blockObject->id());
     delete kvStateCacheBlockBuilder;
   }
   evictedNodeData->RecycleSource();
@@ -326,6 +329,7 @@ Status KVStateCacheBuilder::Merge(std::shared_ptr<KVStateCache> kvStateCache) {
   }
 
   this->version = globalCacheBuilder->GetVersion();
+  globalCacheBuilder->Close();
   return Status::OK();
 }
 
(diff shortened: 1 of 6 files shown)


vegetableysm:
The recycle process cannot be placed in the cache manager destructor. Because the client stored in cache manager is provided by the user. We cannot assume the order of destruction of cache manager and client. Additionally, if the client calls disconnect function before the manager destroys, the recycle process will also fail. So I add a Close function to recycle memory. Even if user forgets to call the Close function, it will not affect the subsequent use of the cache, only cause a memory leak.

dashanji (approved):

modules/llm-cache/ds/kv_state_cache.cc
```diff
@@ -259,6 +259,9 @@ void KVStateCacheBuilder::Delete(std::shared_ptr<NodeData> evictedNodeData) {
   // delete (DataWrapper*) evictedNodeData->nodeData;
   if (evictedNodeData->cleanTreeData) {
     this->rootTree->ClearSubtreeData(treeData);
+    std::shared_ptr<Object> blockObject =
+        kvStateCacheBlockBuilder->_Seal(client);
+    client.DelData(blockObject->id());
```
sighingnow, line 264: A `RETURN_ON_ERROR()` might be needed as `client.DelData` may fail and returns a status.

modules/llm-cache/ds/kv_state_cache.cc
```diff
@@ -403,4 +407,20 @@ KVStateCacheBuilder::~KVStateCacheBuilder() {
...
+  for (auto iter = subTreeDataSet.begin(); iter != subTreeDataSet.end();
+       ++iter) {
+    TreeData* treeData = reinterpret_cast<TreeData*>(*iter);
+    if (treeData->isPtr == true &&
+        treeData->kvStateCacheBlockBuilder != nullptr) {
+      std::shared_ptr<Object> object =
+          reinterpret_cast<KVStateCacheBlockBuilder*>(
+              treeData->kvStateCacheBlockBuilder)
+              ->_Seal(client);
+      client.DelData(object->id());
```
sighingnow, line 421: Keep in mind that everything could fail.

PR #1824, opened 2024-03-14
Bump up golint action to v4 and not specify a version

dashanji:
What do these changes do?
-------------------------

As titled.


dashanji:
Close via https://github.com/v6d-io/v6d/issues/1825

PR #1829, opened 2024-03-15
Fix bug of refcnt map.

### LINKED ISSUE #1828: Add block object id contained in global cache object to refcnt map
Describe your problem
---------------------

Currently refcnt map will only store the refcnt of block object held by clients. Refcnt object may delete the block object  (refcnt may dec to 0) from the global cache object when all clients exit (or, more generally, when a client that has just updated the global object and exits). If another client pull the global object, it may get a block object that is already deleted by refcnt map.

vegetableysm:
What do these changes do?
-------------------------

As tittled.
Refer issue.

Related issue number
--------------------

Fixes #1828

### DIFF
--- a/modules/llm-cache/tests/refcnt_map_test.cc
+++ b/modules/llm-cache/tests/refcnt_map_test.cc
@@ -212,6 +212,9 @@ Status checkRefCnt(std::string ipc_socket) {
           client[0].FetchAndGetObject(globalCacheObjectID));
   kvStateCache->GetCurrentBlockIDSet(blockIDSetToAdd);
   refcnt_map->IncSetRefcnt(blockIDSetToAdd);
+  if (kvStateCache->id() != globalCacheObjectID) {
+    client[0].DelData(kvStateCache->id());
+  }
 
   LOG(INFO) << "Prepare refcnt done";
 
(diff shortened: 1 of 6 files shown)


modules/llm-cache/tests/refcnt_map_test.cc
```diff
@@ -203,6 +203,19 @@ Status checkRefCnt(std::string ipc_socket) {
...
+  ObjectID globalCacheObjectID;
+  blockIDSetToAdd.clear();
+  VINEYARD_CHECK_OK(client[0].GetName(llmCacheObjectName, globalCacheObjectID));
+  std::shared_ptr<KVStateCache> kvStateCache =
+      std::dynamic_pointer_cast<KVStateCache>(
+          client[0].FetchAndGetObject(globalCacheObjectID));
+  kvStateCache->GetCurrentBlockIDSet(blockIDSetToAdd);
+  refcnt_map->IncSetRefcnt(blockIDSetToAdd);
+  if (kvStateCache->id() != globalCacheObjectID) {
+    client[0].DelData(kvStateCache->id());
```
sighingnow, line 216: It might failure...

Keep in mind that any interaction with vineyard may fail, and requires status checking.

PR #1831, opened 2024-03-17
[k8s]: Remove unnecessary time parsing in reconcile functions

pmokeev:
What do these changes do?
-------------------------

Removes unnecessary `time.ParseDuration("1m")` from `Reconcile` functions in controllers package and replaces it by using standard constant `time.Minute` from `time` package.


sighingnow:
Merged. Thank you @pmokeev ❤️

Issue #1841, opened 2024-03-20
Improve the update/query interface in file_storage.cc


vegetableysm:
Describe your problem
---------------------

Currently, the update/query interface in fs storage is only a temporary implementation that needs to be refined.

PR #1842, opened 2024-03-20
Implement the update api on fs.

dashanji:
What do these changes do?
-------------------------

As titled.

Related issue number
--------------------

Fixes part of https://github.com/v6d-io/v6d/issues/1841

### DIFF
--- a/modules/llm-cache/storage/file_storage.cc
+++ b/modules/llm-cache/storage/file_storage.cc
@@ -13,6 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+#include <filesystem>
 #include <map>
 #include <memory>
 #include <string>
@@ -20,6 +21,7 @@ limitations under the License.
 #include <vector>
 
 #include "common/util/logging.h"
+#include "common/util/status.h"
 #include "llm-cache/storage/file_storage.h"
 
 namespace vineyard {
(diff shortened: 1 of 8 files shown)


modules/llm-cache/storage/file_storage.cc
```diff
@@ -39,78 +41,48 @@ namespace vineyard {
...
+  std::vector<std::string> pathList;
+  std::string dir = rootPath;
+  std::string path;
+  int tokenSize;
+
+  VINEYARD_ASSERT(batchSize > 0);
+  // check if the root path ends with '/'
+  if (dir.back() != '/') {
+    dir += "/";
+  }
```
vegetableysm, line 53: Use std::regex_replace(path, std::regex("/+"), "/") to deal with the final path may be better.

modules/llm-cache/storage/file_storage.cc
```diff
@@ -39,78 +41,48 @@ namespace vineyard {
...
-        for (int k = 0; k < kvStateSize; k++) {
-          data += std::to_string((
-                      reinterpret_cast<uint8_t*>(kvStateList[i * batchSize + j]
-                                                     .find(currentLayer)
-                                                     ->second.first.data))[k]) +
-                  " ";
+    std::filesystem::path filePath(dir + pathList[i]);
+    RETURN_ON_ERROR(Mkdir(filePath.parent_path().string()));
+    RETURN_ON_ERROR(Open(filePath.string(), fd, FileOperationType::WRITE));
+    RETURN_ON_ERROR(Seek(fd, 0));
```
vegetableysm, line 66: Currently, when open a file, the read/write pointer is at the begin of the file.

modules/llm-cache/storage/file_storage.cc
vegetableysm: e.g.  user the path as rootPath + "/" + pathList[i]
Then, use the std::regex_replace to replace the multi "/" to "/"

vegetableysm (approved):

modules/llm-cache/storage/file_storage.cc
dashanji: Make sense to me. Thanks

modules/llm-cache/storage/file_storage.cc
dashanji: Based on our discussion, the filesystem api will do the regex itself. E,g, `/vineyard/llm/cache` == `/vineyard////llm//cache`

Issue #1859, opened 2024-04-09
Modifications and memory occupation in vineyard torch module


dashanji:
Describe your problem
---------------------

Thanks for @TrafalgarZZZ reporting. There are two issues when testing vineyard torch module.

#### Reproduce
```python
import safetensors
import safetensors.torch

import vineyard
import vineyard.contrib.ml.torch as vineyard_torch

with open("/mnt/stable-diffusion-models/Stable-diffusion/v1-5-pruned-emaonly-1.safetensors", 'rb') as f:
   state_dict = safetensors.torch.load(f.read())

print(state_dict)

client = vineyard.connect("/tmp/vineyard_test.sock")
with vineyard_torch.torch_context(client):
   client.put(state_dict)

print(state_dict)
```

#### Original state_dict
```bash
...
'cond_stage_model.transformer.text_model.encoder.layers.0.mlp.fc1.weight': tensor([[ 0.0402,  0.0049,  0.0031,  ...,  0.0076, -0.0040, -0.0004],
         [ 0.0320, -0.0247,  0.0270,  ...,  0.0014, -0.0266, -0.0196],
         [-0.0072,  0.0229,  0.0050,  ..., -0.0068, -0.0446, -0.0313],
         ...,
         [ 0.0280, -0.0149,  0.0136,  ...,  0.0182, -0.0120, -0.0161],
         [ 0.0343, -0.0128, -0.0234,  ...,  0.0229, -0.0218,  0.0272],
         [ 0.0184,  0.0124,  0.0135,  ..., -0.0094,  0.0302, -0.0117]]),
 ...}
```

#### state_dict after put
```bash
...
[...]

PR #1860, opened 2024-04-09
Fix the modifications and memory occupation in vineyard torch module.

dashanji:
What do these changes do?
-------------------------

As titled.

Actually, I don't know if simply adding a `value.copy()` is a good way to solve this problem. Could you please take a look at this? @sighingnow 

Related issue number
--------------------

Fixes https://github.com/v6d-io/v6d/issues/1859


dashanji:
> You should make a shallow copy of the state_dict, rather than copy the tensors...

Thanks for the suggestions.

PR #1862, opened 2024-04-10
Add gc for file storage.

### LINKED ISSUE #1861: Add file storage gc for llm module
Describe your problem
---------------------

As tittled. Use timestamp / access time to delete infrequently accessed files

vegetableysm:
What do these changes do?
-------------------------

- Add file storage gc for each client.
- Add global storage to scan dir and delete outdated files.
- Fix bugs in Update when delete files that do not meet requirements.

Related issue number
--------------------

Fixes #1861


modules/llm-cache/storage/file_storage.cc
```diff
@@ -189,15 +199,24 @@ Status FileStorage::Update(
...
+      if (createFileSet.find(this->rootPath + pathList[i]) !=
+          createFileSet.end()) {
+        TouchFile(this->rootPath + pathList[i]);
+        gcList.push_back(this->rootPath + pathList[i]);
+      }
+    } else {
+      lock.~lock_guard();
+      break;
     }
   }
```
sighingnow, line 216: Use lexical scope rather than explicit dtor, i.e.,

```suggestion
  {
    std::lock_guard<std::mutex> lock(gcMutex);
    int j = 0;
    for (size_t i = 0; i < pathList.size(); i++) {
      if (pathIndexMap.find(i) != pathIndexMap.end()) {
        j += 1;
        if (createFileSet.find(this->rootPath + pathList[i]) !=
            createFileSet.end()) {
          TouchFile(this->rootPath + pathList[i]);
          gcList.push_back(this->rootPath + pathList[i]);
        }
      } else {
        break;
      }
    }
  }
```

modules/llm-cache/ds/kv_state_cache_manager.cc
```diff
@@ -400,12 +404,25 @@ Status KVStateCacheManager::ClearGlobalCache(Client& client,
...
 
-void KVStateCacheManager::Close() { storage->CloseCache(); }
+Status KVStateCacheManager::StartGlobalGCThread() {
+  if (this->type != CacheType::FILE) {
+    return Status::Invalid("Only support for file storage");
+  }
+
+  std::dynamic_pointer_cast<FileStorage>(storage)->StartGlobalGCThread();
+  return Status::OK();
+}
```
sighingnow, line 425: It is a bit tedious (and misleading) to have a new config for storage type, imagining you have a manager whose type is blob but the storage is file storage (the current code has no way to prevent such a combination).

You would better implement `StartGlobalGCThread` to blob storage as well (as an empty function) and here you could just call `storage->StartGlobalGCThread()`.

modules/llm-cache/tests/kv_state_cache_local_file_test.cc
```diff
@@ -172,7 +173,14 @@ void inference(std::shared_ptr<KVStateCacheManager>& kv_state_cache_manager,
   }
 }
 
-void threadFunc() {
+void checkFilesNotExist(std::string dir) {
+  for (auto it = std::filesystem::recursive_directory_iterator(dir);
+       it != std::filesystem::recursive_directory_iterator(); ++it) {
+    VINEYARD_ASSERT(!std::filesystem::is_regular_file(*it));
```
sighingnow, line 179: I suggest no `std::filesystem` as it requires C++17, which makes our code cannot be built without such a new C++ compiler.

We have included a thirdparty one in v6d repo, try grep `ghc::filesystem` in the codebase for how to use it.

sighingnow (changes requested):

modules/llm-cache/storage/file_storage.cc
vegetableysm: OK, I understand.

modules/llm-cache/tests/kv_state_cache_local_file_test.cc
vegetableysm: OK

modules/llm-cache/ds/kv_state_cache_manager.cc
vegetableysm: OK

modules/llm-cache/ds/config.h
```diff
@@ -53,16 +53,29 @@ struct FileCacheConfig : public KVCacheConfig {
...
+  int globalTTL;         // second
 
   FileCacheConfig(int tensorByte = 10, int cacheCapacity = 10, int layer = 1,
                   int batchSize = 4, int splitNumber = 2,
                   std::string root = "/tmp/llm_cache/",
-                  FilesystemType filesystemType = LOCAL)
+                  FilesystemType filesystemType = LOCAL,
+                  int clientGCInterval = 3, int ttl = 5,
+                  bool enbaleGlobalGC = false, int globalGCInterval = 30,
+                  int globalTTL = 50)
```
sighingnow, line 68: TTL is too short. You may use 30 minutes and 3 hours for local and global, respectively.

As well as the interval.

sighingnow (approved):
Could you please add related unit tests?

The test case can assert:

- if the file can be correctly deleted
- make sure files that doesn't expire won't be deleted
- check if global GC work after the process/kv-cache manager restarts.

vegetableysm:
> Could you please add related unit tests?
> 
> The test case can assert:
> 
> * if the file can be correctly deleted
> * make sure files that doesn't expire won't be deleted
> * check if global GC work after the process/kv-cache manager restarts.

OK, I'll add it in this pr.

modules/llm-cache/ds/config.h
vegetableysm: Done

sighingnow (approved):

PR #1863, opened 2024-04-11
Refine the vineyard client wrapper to handle the corner case and avoid the block of IO

dashanji:
What do these changes do?
-------------------------

Related issue number
--------------------

Fixes #issue number

### DIFF
--- a/python/vineyard/core/client.py
+++ b/python/vineyard/core/client.py
@@ -19,6 +19,8 @@
 import contextlib
 import os
 import warnings
+from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import as_completed
 from typing import Any
 from typing import Dict
 from typing import List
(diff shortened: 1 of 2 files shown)


python/vineyard/core/client.py
```diff
@@ -675,7 +687,10 @@ def put(
         try:
             return put(self, value, builder, persist, name, **kwargs)
         except NotEnoughMemoryException as exec:
+            print("called _find_the_most_available_memory_instance...")
             instance_id = self._find_the_most_available_memory_instance()
+            print("after called _find_the_most_available_memory_instance...")
+            previous_compression_state = self.compression
```
sighingnow, line 693: Delete these two debug logs.

python/vineyard/core/client.py
```diff
@@ -622,27 +635,33 @@ def _find_the_most_available_memory_instance(self) -> int:
             int: The instance id with the most available memory.
             if only have one instance, return -1
         """
+        os.environ['VINEYARD_RPC_SKIP_RETRY'] = '1'
```
sighingnow, line 638: Use `with vineyard.envvars()....` to avoid the global variable been populated.

dashanji: Make sense. Thanks.

python/vineyard/core/client.py
dashanji: Ok.

PR #1869, opened 2024-04-19
Fixes the permission & build args for vineyard-fuse image

sighingnow:
Fixes https://github.com/v6d-io/v6d/actions/runs/8731399581/job/23978461191


docker/Makefile
```diff
@@ -126,6 +126,7 @@ build-python-dev:
 vineyard-fluid-fuse:
 	docker build ./vineyard-fluid-fuse \
 		-t $(VINEYARD_FLUID_FUSE_REGISTRY)/$(VINEYARD_FLUID_FUSE_IMAGE):$(VINEYARD_FLUID_FUSE_TAG) \
+		--platform linux/$(ARCH) \
```
dashanji, line 129: Thanks for pointing it out.

Issue #1890, opened 2024-05-15
Repeated GetBuffer calls by a client instance for a single get operation.


ysbqiaqia:
Hi vineyard developers,

Recently, we reaserch vineyard flow, I found that when I do one single get operation in python script, client executed multiple GetBuffer requests. I added a print log at command_t::GET_BUFFERS_REQUEST state in the processMessage state mechine （at line **259** in the **socket_server.cc**）. The printed log show that at least three invocations of the GetBuffer request. Here is the log: from vineyardd:
```
// log position:
  } else if (cmd == command_t::GET_BUFFERS_REQUEST) {
    printf("[\033[33m%s\033[0m] %s %s doGetBuffers\n", my_time_stamp().c_str(), __FILE__, __FUNCTION__);
    return doGetBuffers(root);
}
...
} else if (cmd == command_t::GET_DATA_REQUEST) {
    printf("[\033[33m%s\033[0m] %s %s doGetData\n", my_time_stamp().c_str(), __FILE__, __FUNCTION__, __LINE__);
    return doGetData(root);
  } 

[2024-05-15 18:03:36.939] v6d/src/server/async/socket_server.cc processMessage doGetData
[2024-05-15 18:03:36.941] v6d/src/server/async/socket_server.cc processMessage doGetBuffers
[2024-05-15 18:03:36.942] v6d/src/server/async/socket_server.cc processMessage doGetData
[2024-05-15 18:03:36.945] v6d/src/server/async/socket_server.cc processMessage doGetBuffers
[2024-05-15 18:03:36.948] v6d/src/server/async/socket_server.cc processMessage doGetData
[2024-05-15 18:03:36.949] v6d/src/server/async/socket_server.cc processMessage doGetBuffers
```
Test Demo:
```  
#-------------------- ipc_put.py
import vineyard
import numpy as np
import os, sys

if __name__ == '__main__':
    if len(sys.argv) != 2:
        print("using : python main.py size(KB unit)")
        exit(-1)
    
    client = vineyard.connect('/var/run/vineyard.sock')
    b4 = 2 * np.ones((128,int(sys.argv[1])),dtype=int)
    obj_id = client.put(b4)
    print(obj_id)


#--------------------  ipc_get.py
import vineyard
import numpy as np
import os, sys
if __name__ == '__main__':
    client = vineyard.connect('/var/run/vineyard.sock')
    if len(sys.argv) != 2:
        print("using : python main.py obj-id")
        exit(-1)
[...]

PR #1893, opened 2024-05-21
Construct object from meta to fix the repeated getbuffers.

dashanji:
What do these changes do?
-------------------------

As titled.

Related issue number
--------------------

Fixes https://github.com/v6d-io/v6d/issues/1890


sighingnow (changes requested):
Refer to https://pybind11.readthedocs.io/en/stable/advanced/classes.html for how to add a `__init__` method to pybind objects.

PR #1903, opened 2024-06-03
Add rdma support

### LINKED ISSUE #1895: Add rdma lib for vineyard
As tittled

vegetableysm:
What do these changes do?
-------------------------
Done:
- Add rdma lib
- Integration with rdma lib
- Refactor code
- **Roll back** GenerateBlobID. I will implement create objectID from virtual address in the future. Directly use virtual address and instance id as the blob ID will trigger error when vineyardd triggers spill. Because the GenerateBlobID function will assign a assigned ObjectID to a new object when put a cold object to disk.(Because there is no change of the object virtual address and the instance id )
- Split big size data transfer with rdma into multiple small size data transfer.
- Register memory when use RDMA to get/create remote object.(To solve the problem of mr's limit.)

Todo:
- Compress/Decompress

Related issue number
--------------------

Fixes #1895


CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ option(BUILD_VINEYARD_FUSE "Enable vineyard's fuse support" OFF)
 option(BUILD_VINEYARD_FUSE_PARQUET "Enable vineyard's fuse parquet support" OFF)
 option(BUILD_VINEYARD_HOSSEINMOEIN_DATAFRAME "Enable hosseinmoein dataframe support" OFF)
 option(BUILD_VINEYARD_LLM_CACHE "Enable kv-state cache support" ON)
+option(BUILD_VINEYARD_RDMA "Enable RDMA support" OFF)
```
sighingnow, line 66: Build RDMA support by default.

sighingnow (changes requested):
Chan

CMakeLists.txt
vegetableysm: Libfabric may not be on the build environment. Make it as a submodule?

dashanji:
Hi @vegetableysm, could you please add a doc about how to use rdma?

vegetableysm:
> Hi @vegetableysm, could you please add a doc about how to use rdma?

OK!

PR #1959, opened 2024-07-25
Prevents objects from being deleted during transfer.

### LINKED ISSUE #1927: If the object is deleted when the client migrates it, the client will read the invalid value
Describe your problem
---------------------

As tittled, example:
1. IPCClient create a object (object id is id1)
2. RPCClient get object with id1, and get blob with object meta
3. RPCClient call getRemoteBlob to get blob
4. Before Vineyardd send blob data, vineyard sleep 5 second.
5. IPCClient call delData to delete the object and create a new object with new data
6. Vineyard wakes, and sends buffer.
7. RPCClient get blob data.

I think the rpc client should get the error status because the object has been deleted. Or vineyardd should refuse to delete the object being migrated during migration. However, rpc client get the new data from vineyardd.

Screen shot
Create object with data {1.0, 7.0, 3.0, 4.0, 2.0} and delete it, then create object with data {11, 23, 34, 45, 56}
image
Check will failed because rpc client use the old blob id and read new data with {11, 23, 34, 45, 56}

Test code has been pushed at https://github.com/vegetableysm/v6d/tree/test-rpc-with-del

vegetableysm:
What do these changes do?
-------------------------

As tittled.

Related issue number
--------------------

Fixes #1927

### DIFF
--- a/src/client/rpc_client.cc
+++ b/src/client/rpc_client.cc
@@ -747,6 +747,7 @@ Status RPCClient::GetRemoteBlob(const ObjectID& id, const bool unsafe,
   std::vector<int> fd_sent;
 
   std::string message_out;
+  RDMABlobScopeGuard rdmaBlobScopeGuard;
   if (rdma_connected_) {
     WriteGetRemoteBuffersRequest(std::set<ObjectID>{id}, unsafe, false, true,
                                  message_out);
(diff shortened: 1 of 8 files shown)


dashanji:
@vegetableysm Could you please add a test to valid this?

Issue #1972, opened 2024-08-06
Evict the vineyard object but not store in the disk


dashanji:
Describe your problem
---------------------

In the llm kv cache scenery, we need to spill these objects with access time. It's useful to leverage the external storage to increase the token hit rate.

PR #1983, opened 2024-08-12
Support to spill objects with LRU strategy by vineyardd itself.

dashanji:
What do these changes do?
-------------------------

 As titled.

Related issue number
--------------------

Fixes https://github.com/v6d-io/v6d/issues/1972

### DIFF
--- a/python/client.cc
+++ b/python/client.cc
@@ -826,6 +826,22 @@ void bind_client(py::module& mod) {
              throw_on_error(self->Fork(*client));
              return client;
            })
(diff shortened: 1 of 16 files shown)


vegetableysm:
The commit 524157e86be1e5bfa3ab1df06b3588429e544b38 fix:
- Dead lock
- If a thread get a meta of a spilled object, then next thread reload it from disk. The pre thread will find out that the payload pointer is nullptr and try to reload it again.

Issue #2003, opened 2024-10-17
name不是数据对象的唯一标识问题


1076881851:
以name作为数据对象的唯一标识来存取数据，这种场景用户理解name是唯一标识，只存一份数据。但是当前以同一个name put两份数据，两份数据都会存进vineyard数据库里，元数据里的name是一样的；通过name取的是最新的一份数据，请问vineyard的设计是如何考虑的？老的数据是不是相当于残留了？
![image](https://github.com/user-attachments/assets/f4e7eacb-7ea8-44a1-9942-8984c2adb718)
![image](https://github.com/user-attachments/assets/5d7aad91-dfad-4778-bd81-c769328530f8)

dashanji:
你可以通过object id拿到之前的数据，如果必须要用name对应一个object的时候，需要确保name不会被覆盖。另外你预期覆盖后是个什么行为比较好呢？

Daniel-blue:
1. When operating an object through name, currently it appears that name is associated with the latest object; So what is the purpose of retaining the name in the original object, for historical tracking?
2. Can it be assumed that a user no longer needs an old object with the same name when writing a new object using the same name, can it be automatically deleted, or can a prompt be returned to the user for decision?

Daniel-blue:
1. At present, when using the same name to write an object, the name will be associated with the latest written object, but existing objects will also retain that name Why not remove this name from existing external objects? What is the purpose of retaining this name on an existing object? Is it because of immutable data?

2. Can a user assume that they no longer need an existing object with the same name when writing a new object using that name? Can the existing object be automatically deleted, or can a prompt be returned to the user for decision? Or rather, due to immutable data, no deletion or other operations are performed; Only new objects can be added and associated with that name

dashanji:
@Daniel-blue Thanks for the detail. We can validate whether the name exists and make sure the object-name mapping not to be overwritten.

PR #2010, opened 2024-10-31
Improve the name mechanism of vineyard objects.

dashanji:
What do these changes do?
-------------------------

* Make sure one name is associated with one vineyard object.
* Fix the name can't be all digits.
* Support delete vineyard object by name.

Related issue number
--------------------

Fixes https://github.com/v6d-io/v6d/issues/2003

### DIFF
--- a/src/server/server/vineyard_server.cc
+++ b/src/server/server/vineyard_server.cc
@@ -57,6 +57,14 @@ namespace vineyard {
   } while (0)
 #endif  // ENSURE_VINEYARDD_READY
 
+// Helper function to adjust the name if it's all digits.
+std::string AdjustName(const std::string& name) {
+  if (std::regex_match(name, std::regex("[0-9]+$"))) {
+    return "_" + name;
+  }
+  return name;
+}
+
 bool DeferredReq::Alive() const { return alive_fn_(); }
 
 bool DeferredReq::TestThenCall(const json& meta) const {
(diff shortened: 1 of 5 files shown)


src/server/server/vineyard_server.cc
```diff
@@ -57,6 +57,14 @@ namespace vineyard {
   } while (0)
 #endif  // ENSURE_VINEYARDD_READY
 
+// Helper function to adjust the name if it's all digits.
+std::string AdjustName(const std::string& name) {
+  if (std::regex_match(name, std::regex("[0-9]+$"))) {
```
sighingnow, line 62: Use https://en.cppreference.com/w/cpp/string/byte/isdigit. Regex is expensive.

src/server/server/vineyard_server.cc
```diff
@@ -871,13 +879,22 @@ Status VineyardServer::PutName(const ObjectID object_id,
                                const std::string& name, callback_t<> callback) {
   ENSURE_VINEYARDD_READY();
   auto self(shared_from_this());
+  // if the name is all digits, prefix it with '_'
+  // as the nlohmann/json can't convert '/names/1234: 12345' to 'names:{"1234",
+  // "12345"}'
+  std::string new_name = AdjustName(name);
+  std::cout << "new_name is: " << new_name << std::endl;
```
sighingnow, line 886: Remove those debugging logs.

sighingnow (changes requested):

PR #2052, opened 2025-07-26
Fix build for Apache Arrow 21.0.0

amoeba:
What do these changes do?
-------------------------

Fix v6d when being built under Apache Arrow 21.

DCHECK_EQ was hidden in Arrow 21, see https://github.com/apache/arrow/issues/46011. ARROW_DCHECK_GE should be used instead.

Related issue number
--------------------

No.


dashanji:
@amoeba Thanks!!!!

PR #2054, opened 2025-07-30
Created a yaml file for HPA

### LINKED ISSUE #1302: Support autoscaling of vineyard cluster on the kubernetes.
Describe your problem
---------------------

Although we have the spill/checkpoint mechanism to handle the scenarios out of memory locally, the existing mechanism is not suitable on the kubernetes, so we hope to introduce the [Horizontal Pod Autoscaling](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) mechanism to implement the autoscaling of the vineyard cluster on the kubernetes. 
The issue mainly includes the following tasks.

- [ ] Find a proper watermark of memory usage to scale out or scale in.
- [ ] Expose the metric of memory usage to the vineyard operator.
- [ ] Create a HorizontalPodAutoscaler for controlling the scaling of the vineyard cluster.

Rajdeep1311:
What do these changes do?
-------------------------

Created a yaml file for Horizontal Pod Autoscaler (HPA) that scales up and down the vineyard cluster according to the memory usage by the cluster

Related issue number
--------------------

Fixes #1302


Rajdeep1311:
@dashanji Sir am I supposed to include the HPA in such a way that when the vineyard cluster is deployed, the hpa also runs. And include the values of hpa in values.yaml?

Rajdeep1311:
> okay， after that you would better to add a doc to show a complete example.

@dashanji Sure sir

Rajdeep1311:
@dashanji Sir could you please review the PR?

dashanji:
Hi @Rajdeep1311, I think you may misunderstand the target resource defined in the `scaleTargetRef`.  It should be the `vineyard cluster` rather than the `vineyard operator`.
You can refer to the doc (https://v6d.io/tutorials/kubernetes/using-vineyard-operator.html) to create a vineyard cluster and get some insights from it.

Rajdeep1311:
> Hi @Rajdeep1311, I think you may misunderstand the target resource defined in the `scaleTargetRef`. It should be the `vineyard cluster` rather than the `vineyard operator`. You can refer to the doc (https://v6d.io/tutorials/kubernetes/using-vineyard-operator.html) to create a vineyard cluster and get some insights from it.

Sir @dashanji , will the `scaleTargetRef` be the vineyard cluster created as a custom resource?

dashanji:
> 
> Sir @dashanji , will the `scaleTargetRef` be the vineyard cluster created as a custom resource?

Yes.

Rajdeep1311:
> > Sir, I am facing a problem, the vineyardd custom resource (CR) isn't populating the `.status.selector`, I need that field so that the HPA can patch up with the cluster. As of now, I'm getting a ` selector is required` error from the HPA that I am creating. I need the selectors so that HPA can identify the vineyard cluster.
> 
> You can use the vineyard deployments' selector rather than the vineyardd CR.

Yeah, sir, I will get it from the deployment, but I think I was not able to communicate my question properly.
So, will I point the scaleTargetRef to the deployment created by the vineyard custom resource? But we need to scale the vineyard custom resource, right, Sir, not the deployment created by the custom resource? @dashanji

If we need to scale the custom resource itself, then the HPA needs the status.selector field to be populated on creation of the CR, we cannot provide that manually, sir, but it is not happening. The HPA needs status.selector to identify the CR.

PR #2066, opened 2026-01-20
Support building with glog 0.7+

cho-m:
What do these changes do?
-------------------------

Support compilation with glog 0.7.0+.

Related issue number
--------------------

n/a


cho-m:
Closing. Originally created for Homebrew but we have decided to deprecate and ultimately remove Vineyard due to lack of response.

PR #2073, opened 2026-08-09
Reclaim created-but-unsealed blobs when the client connection is closed

### LINKED ISSUE #2072: Created-but-unsealed blobs leak permanently when a client disconnects before sealing (v0.24.2)
> **Note:** this issue is filed by an AI agent on behalf of its operator (@mpups), who has verified the diagnosis and can follow up on questions.

Describe your problem
---------------------

**Version:** v0.24.2 (the `vineyardd` binary bundled in the `vineyard` Python wheel), Linux x86_64.

When a client creates a buffer with `create_blob` / `CreateBuffer` and then dies (or otherwise disconnects) **before sealing it**, the allocation is never reclaimed by the daemon. The bytes stay counted as "in use" forever — across the dying client's disconnect, across later client connect/disconnect cycles, until the daemon itself is restarted.

Because the blob was never written, no shared-memory pages are ever touched: the daemon's RSS stays at a few MiB and `/dev/shm` shows nothing (the arena is `memfd_create(2)`-backed). The leak is pure allocator bookkeeping, which makes it invisible to the usual memory-inspection tools.

Minimal reproduction
--------------------

```bash
mkdir -p /tmp/vy-repro
vineyardd --socket=/tmp/vy-repro/vineyard.sock --size=1Gi --meta=local --rpc=false &
```

Client 1 — create a 400 MiB blob and die without sealing:

```python
# leak_client.py
import os
import vineyard

client = vineyard.connect('/tmp/vy-repro/vineyard.sock')
buf = client.create_blob(419430400)   # 400 MiB, never sealed
os._exit(0)                           # die abruptly, no seal, no clean shutdown
```

Client 2 — fresh process, attempt an oversized put to read the usage accounting:

```python
# check_client.py
import numpy as np
import vineyard

client = vineyard.connect('/tmp/vy-repro/vineyard.sock')
data = np.zeros(700 * 1024 * 1024, dtype=np.uint8)  # 700 MiB > 1 GiB - 400 MiB
client.put(data)
```

Observed output from client 2 (immediately after client 1 dies, and identically 10+ seconds later, after further connect/disconnect cycles):

```
Not enough memory: IPC error at /work/v6d/src/common/util/protocols.cc:351:
Failed to allocate memory of size 734003200, total available memory size are 1073741824,
and 419430400 are already in use
```

The "already in use" figure is exactly the unsealed blob from the dead client. Meanwhile the daemon's `VmRSS` is ~2 MiB — no pages were ever committed.

Expected behavior
-----------------

When a connection closes (gracefully or not), the daemon's connection-teardown path (`ReleaseConnection` / socket-close handling) should drop or roll back any buffers that were created by that connection and never sealed, returning their reservation to the pool — analogous to how dependencies/streams are cleaned up on client death.

Impact
------

In production we cache large model weights (~35 GiB blobs) in vineyard. A client pod that crash-loops mid-put — dies between `create_buffer` and `seal`, e.g. OOM-killed or evicted while filling the buffer — leaks its full blob size on every attempt. A handful of restarts is enough to wedge the entire pool at capacity: every subsequent create is refused with the "Not enough memory" error above, while the daemon looks completely healthy (tiny RSS, empty `/dev/shm`, responsive IPC). With spilling disabled there is no reclamation path at all short of restarting `vineyardd` and losing the whole cache. We observed exactly this: the pool frozen with 5 leaked copies of the same 35 GiB weights blob booked as "in use" with zero corresponding resident pages.

A daemon-side fix (reclaim unsealed buffers on connection teardown) would remove this failure mode entirely; a `vineyardctl`/API way to list and drop unsealed allocations would also help as a mitigation.

mpups:
> **Note:** this pull request was prepared by an AI agent on behalf of its operator (@mpups), who has reviewed the change and can follow up on questions.

What do these changes do?
-------------------------

When a client creates a buffer (`CreateBuffer` / `CreateBuffers` / `CreateDiskBuffer`) and disconnects — gracefully or by crashing — before sealing it, the allocation stayed booked as "in use" forever. The dependency tracker only learns about a blob at seal/get time, so the connection-teardown path (`ReleaseConnection`) had nothing to release for unsealed blobs, and the reservation leaked until the daemon was restarted. A crash-looping client that dies between `create_buffer` and `seal` can wedge the whole pool at capacity while the daemon looks healthy (tiny RSS, responsive IPC).

This PR makes the daemon reclaim such blobs on connection teardown:

- `SocketConnection` now tracks the blobs it created that haven't been sealed yet (`unsealed_blobs_`, alongside the existing `associated_streams_` cleanup idiom). Entries are removed on `SealBlob` and `DropBuffer`.
- On `Stop()`, any tracked blob that is *still* unsealed is deleted via a new `BulkStoreBase::DeleteUnsealed()`, which shares its implementation with `Delete()` but skips (atomically, inside the cuckoohash `erase_fn` predicate) blobs that have been sealed in the meantime — so a blob sealed by another connection after being shared is unaffected, and non-owner blobs (e.g. moved between sessions) are left alone as before.

Out of scope (unchanged behavior): GPU buffers (`CreateGPUBuffer`, separate `DeleteGPU` path), the plasma store, and remote buffers over RPC (`CreateRemoteBuffer` already seals eagerly and cleans up on receive failure).

**Testing:** new `test/unsealed_blob_leak_test.cc` (registered in `test/runner.py`) checks that (1) an unsealed blob's reservation returns to baseline after its creator disconnects, (2) a blob aborted before disconnect doesn't double-free, and (3) sealed blobs survive their creator's disconnect and remain readable. Verified locally on ubuntu 22.04: the new test fails against an unpatched `vineyardd` (reservation never returns to baseline) and passes with the fix; `get_blob_test`, `mutable_blob_test`, `release_test`, and `delete_test` all still pass.

Related issue number
--------------------

Fixes #2072


mpups:
@coderabbitai review

